PR #2, opened 2017-07-29
Major overhaul to port manubot-rootstock functionality

dhimmel:
This PR aims to move much of the Python code from manubot-rootstock to this package.


dhimmel:
@agitter big changes on the way. Packaging lot's of the manubot-rootstock python code here.

Another goal which I haven't mentioned elsewhere is to cache citation metadata requests at the earlier possible stage. Therefore if the processing code changes, we don't need to flush the entire cache every time. This will likely use `requests-cache`. Still looking at the best way to use it in https://github.com/reclosedev/requests-cache/issues/96.

This PR will change the referencing format to be pandoc compatible as per https://github.com/greenelab/manubot-rootstock/issues/2#issuecomment-312153192. If there are major breaking changes that you've wanted but previously deemed too destructive, now's the time to suggest them.

agitter:
Structuring this as a distinct package is a great idea.  It will help separate the content from the build system and minimize our need to merge `manubot-rootstock` changes into a cloned repository.

After changing citations to be pandoc compatible, will we still use the previous syntax such as `@doi:10.1371/journal.pone.0032235 and @url:https://github.com/greenelab/manubot` or switch to the tag-like syntax in the linked comment, such as `@smith04`?

I haven't looked at the code changes closely.  Did you reimplement the essential parts of `arxiv2bib` on your own?

dhimmel:
> Did you reimplement the essential parts of arxiv2bib on your own?

Yes. No more bibtex and the arXiv citeproc should now be more extensive.

> After changing citations to be pandoc compatible, will we still use the previous syntax such as `@doi:10.1371/journal.pone.0032235 and @url:https://github.com/greenelab/manubot` 

Yes, but you will need a semicolon to separate multiple references:

```
[@doi:10.1371/journal.pone.0032235; @url:https://github.com/greenelab/manubot]
```

But you will be able to do things like:

```
@doi:10.1371/journal.pone.0032235 says blah
[@doi:10.1371/journal.pone.0032235, p 33-35; @url:https://github.com/greenelab/manubot].
```

See [this](https://github.com/dhimmel/manubot/blob/f818e294a2cff7ff4f0aee56d8d19a1f6940beb7/manubot/citations.py#L9-L14). Currently, I've deviated from pandoc style in 1 important way. References can end in slash. This is because so many URLs end in slash. We don't have to allow this (you can use tags for citations that aren't valid pandoc citation syntax).

agitter:
> Currently, I've deviated from pandoc style in 1 important way.

I agree with that choice.  URLs ending with slash are too common a case to force the use of tags.

dhimmel:
Manubot is now a command line utility. It now requires a specific set of files to be in the content directory. Process the example-manuscript in the PR like:

```
manubot \
  --content-directory=tests/example-manuscript/content \
  --output-directory=tests/example-manuscript/output \
  --log-level=INFO
```

The process should now be more tolerant to manuscript citation errors. Builds will still fail if there are any errors, but the process should complete.

@agitter now's a good time to review.

manubot/manubot.py
```diff
@@ -0,0 +1,269 @@
...
+    logger = logging.getLogger()
+    logger.setLevel(getattr(logging, args.log_level))
+    stream_handler = logging.StreamHandler(stream=sys.stderr)
+    stream_handler.setFormatter(logging.Formatter('## {levelname}\n{message}', style='{'))
+    logger.addHandler(stream_handler)
+
+    prepare_manuscript(args)
+
+    if error_handler.fired:
+        print('Failure: exiting with code 1 due to logged errors')
```
cjw296: I'd probably have this as a log myself ;-)

agitter:
> now's a good time to review.

@dhimmel how thorough of a review would you like?  I'll have to wait until this weekend at the earliest to do a complete line-by-line review.

dhimmel:
> how thorough of a review would you like? I'll have to wait until this weekend at the earliest to do a complete line-by-line review.

Happy to wait. This pull request requires lot's of updates to manuscript repos, so we should review it carefully and make sure we get everything right.

dhimmel:
https://github.com/greenelab/manubot/pull/2/commits/8ca7c86452eccef0cc48eb21d7fa2a91ba5daa99 enables providing additional template variables for jinja2, like:

```
manubot \
  --content-directory=tests/example-manuscript/content \
  --output-directory=tests/example-manuscript/output \
  --template-variables-path=https://github.com/greenelab/scihub/raw/c59273a4b5a91fd905e88662191d2a213497a682/0.configuration.json \
  --log-level=INFO
```

Therefore, you could have your analysis repo generate a bunch of stats, which you refer to by variable in the manuscript.

agitter:
@dhimmel Because there are a lot of changes here, would it be helpful to have one of the contributors from `manubot-rootstock` review this as well?  Even this weekend I may not be able to do a sufficiently deep review on my own.

dhimmel:
> Because there are a lot of changes here, would it be helpful to have one of the contributors from manubot-rootstock review this as well?

It's very hard to review this many changes at once. But @slochower @vsmalladi, your comments are welcome.

We can also merge this PR without extensive review and take the codebase from there with follow up PRs to address any changes people have.

slochower:
I'm less familiar with doing code review through GitHub's interface. Is there a way I could merge the PR on a local clone of the repository and run it through test of various citation styles to make sure all's well?

dhimmel:
@slochower checking out the code from a PR is a great way to review it. Oftentimes, there is no substitute for actually testing code.

You can [checkout this PR](https://help.github.com/articles/checking-out-pull-requests-locally/) into a branch. Let's assume this repo is your `upstream` remote. Then you can do:

```
git fetch upstream pull/2/head:pr-2
git checkout pr-2
```

vsmalladi:
Well start taking a look and see if I notice anything.

slochower:
Me too, sorry didn't get a chance to do it today, but I can take a look early next week if this isn't merged by then.

dhimmel:
> I don't understand the Invalid response from https://data.datacite.org/10.7287%2Fpeerj.preprints.3100v1 error because the URL works in my browser.

Reported upstream at https://github.com/datacite/datacite/issues/187. I'll skip the test for time being.

manubot/manubot.py
```diff
@@ -0,0 +1,317 @@
...
+
+
+def parse_arguments():
+    """
+    Read and process command line arguments.
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--content-directory', type=pathlib.Path, required=True)
+    parser.add_argument('--output-directory', type=pathlib.Path, required=True)
+    parser.add_argument('--template-variables-path',
```
agitter: It would be nice to include an example manuscript that uses the template variables feature, perhaps in `manubot-rootstock` instead of here.

manubot/manuscript.py
```diff
@@ -0,0 +1,64 @@
...
+    # Number of distinct references by type
+    ref_counts = (
+        citation_df
+        .standard_citation
+        .drop_duplicates()
+        .map(lambda x: x.split(':')[0])
+        .pipe(collections.Counter)
+    )
+    ref_counts['total'] = sum(ref_counts.values())
+    stats['reference_counts'] = ref_counts
```
agitter: Do the reference counts and word count get logged at some level?  I may have missed it.

dhimmel: No ATM they only get saved to `variables.json`. Hence they're available to the jinja2 templator / whenever you have access to the output files (anytime besides PR CI builds). Do you think all of variables.tsv should be logged? In what cases do you like seeing the reference and word counts?

manubot/manubot.py
dhimmel: Good point, will add an example manuscript under `tests` for testing `--template-variables-path`

vsmalladi:
I have ran the installation and test code locally and everything seems to be working. I am starting to look line by line at the code.

manubot/manuscript.py
vsmalladi: Would it be possible to have the user set a word count limit that would trigger logging?

agitter: @dhimmel if the manuscript has a word count or reference limit imposed by a journal or conference, logging those values so they can be inspected during PR review is potentially useful.  Those two variables have special significance to me.  I don't think it is necessary to log all of `variables.tsv`.

@vsmalladi I think that setting a word count threshold to trigger logging would be more complicated than logging at an appropriate level.

vsmalladi: @agitter Ya i think logging the word  count and reference limit would be what i would look at. I think the only other thing I would possibly log, would be the number of figures. Since each journal could also have a limit to number of figures.

agitter: @vsmalladi that's a good idea, it could log figure and table counts.  That could come in a later pull request if @dhimmel wants to get this massive one merged.

vsmalladi: @agitter That sound reasonable.

dhimmel:
Okay thanks everyone for the feedback. In the interest of time, I'm going to merge.

PR #8, opened 2017-08-11
Log timezone

dhimmel:
Refs https://github.com/greenelab/manubot/issues/3


dhimmel:
@slochower this logs the timezone and makes our date timezone aware.

PR #22, opened 2018-01-08
Author affiliation numbering

dhimmel:
Generate author affiliation numbering using Manubot. These are exposed as part of the template variables, so that manuscripts can use them. This will assist with alternative frontmatter styles, like the format desired by the Deep Review, see https://github.com/greenelab/deep-review/issues/700.
 

Todos:

- [x] Standardize affiliations format in variables

### DIFF
--- a/manubot/manubot.py
+++ b/manubot/manubot.py
@@ -387,6 +387,10 @@ def main():
     # See https://stackoverflow.com/a/45446664/4651668
     error_handler = errorhandler.ErrorHandler()
 
+    # Log DeprecationWarnings
+    warnings.simplefilter('always', DeprecationWarning)
+    logging.captureWarnings(True)
+
     # Log to stderr
     logger = logging.getLogger()
     stream_handler = logging.StreamHandler(stream=sys.stderr)
--- a/manubot/manubot.py
+++ b/manubot/manubot.py
@@ -193,8 +193,8 @@ def add_author_affiliations(variables):
     for author in variables['authors']:
         name = author['name']
         affiliations = author.get('affiliations', [])
-        affiliations = (affiliations if isinstance(affiliations, list) else
-                        affiliations.split('; '))
+        if not isinstance(affiliations, list):
+            affiliations = affiliations.split('; ')
         for affiliation in affiliations:
             rows.append((name, affiliation))
     if not rows:
--- a/manubot/manubot.py
+++ b/manubot/manubot.py
@@ -182,6 +182,36 @@ def read_jsons(paths):
     return user_variables
 
 
+def add_author_affiliations(variables):
+    """
+    Edit variables to contain numbered author affiliations. Specifically,
+    add a list of affiliation_numbers for each author and add a list of
+    affiliations to the top-level of variables. If no authors have any
+    affiliations, variables is left unmodified.
+    """
+    rows = list()
+    for author in variables['authors']:
+        name = author['name']
+        affiliations = author.get('affiliations', [])
+        affiliations = (affiliations if isinstance(affiliations, list) else
+                        affiliations.split('; '))
+        for affiliation in affiliations:
+            rows.append((name, affiliation))
+    if not rows:
+        return variables
+    affil_map_df = pandas.DataFrame(rows, columns=['name', 'affiliation'])
+    affiliation_df = affil_map_df[['affiliation']].drop_duplicates()
+    affiliation_df['affiliation_number'] = range(1, 1 + len(affiliation_df))
+    affil_map_df = affil_map_df.merge(affiliation_df)
+    name_to_numbers = {name: list(df.affiliation_number) for name, df in
+                       affil_map_df.groupby('name')}
+    for author in variables['authors']:
+        numbers = name_to_numbers.get(author['name'], [])
+        author['affiliation_numbers'] = numbers
+    variables['affiliations'] = affiliation_df.to_dict(orient='records')
+    return variables
+
+
 def get_metadata_and_variables(args):
     """
     Process metadata.yaml and create variables available for jinja2 templating.
@@ -210,6 +240,7 @@ def get_metadata_and_variables(args):
     authors = metadata.pop('author_info', [])
     metadata['author-meta'] = [author['name'] for author in authors]
     variables['authors'] = authors
+    variables = add_author_affiliations(variables)
 
     # Set repository version metadata for CI builds only
     repo_slug = os.getenv('TRAVIS_REPO_SLUG')
--- a/manubot/manubot.py
+++ b/manubot/manubot.py
@@ -203,11 +203,10 @@ def add_author_affiliations(variables):
     affiliation_df = affil_map_df[['affiliation']].drop_duplicates()
     affiliation_df['affiliation_number'] = range(1, 1 + len(affiliation_df))
     affil_map_df = affil_map_df.merge(affiliation_df)
-    name_to_numbers = {name: list(df.affiliation_number) for name, df in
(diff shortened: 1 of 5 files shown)


tests/test_manubot.py
```diff
@@ -67,3 +70,36 @@ def test_read_jsons():
...
+    returned_variables = add_author_affiliations(variables)
+    assert variables is returned_variables
+    assert variables['affiliations'] == [
+        {'affiliation': 'Department of Doe, University of Roe', 'affiliation_number': 1},
+        {'affiliation': 'Peppertea University', 'affiliation_number': 2},
+        {'affiliation': 'Unique University', 'affiliation_number': 3},
+    ]
+    authors = variables['authors']
+    assert authors[0]['affiliation_numbers'] == [1, 2]
+    assert authors[1]['affiliation_numbers'] == [3, 2]
```
dhimmel: Currently the [build fails](https://travis-ci.org/greenelab/manubot/builds/326477396#L1250) because of an ordering issue in pandas. From https://github.com/pandas-dev/pandas/issues/18776#issuecomment-356053890:

pandas-inner-merge-order

When I wrote the code, I intended for the order of the numbers to match the order author affiliations were listed in `metadata.yaml`. However, one could also argue that numbers should always be sorted.

In other words, sorted numbers would force:

> Daniel Himmelstein<sup>1,2</sup>

Is that good or bad. Should author superscripts match the order they input in `metadata.yaml`. If yes, we could have:

> Daniel Himmelstein<sup>2,1</sup>

@agitter what do you think the ideal behavior is here?

manubot/manubot.py
```diff
@@ -182,6 +182,36 @@ def read_jsons(paths):
...
+    add a list of affiliation_numbers for each author and add a list of
+    affiliations to the top-level of variables. If no authors have any
+    affiliations, variables is left unmodified.
+    """
+    rows = list()
+    for author in variables['authors']:
+        name = author['name']
+        affiliations = author.get('affiliations', [])
+        affiliations = (affiliations if isinstance(affiliations, list) else
+                        affiliations.split('; '))
```
dhimmel: I believe now that an authors affiliations should always be a list in `metadata.yaml`. However, to keep backwards compatibility, `;  `-separated affiliations are accepted.

@agitter do you think affiliations should be a list in `metadata.yaml`? If so, I'll update the manubot-rootstock template after this is merged? Do we want to retain backwards compatibility and if so is this the best way?

agitter:
> what do you think the ideal behavior is here?

I think that the affiliation superscripts should always be sorted.  The index would be assigned based on the first author in `metadata.yml` who has that affiliation.  If an author introduces multiple new affiliations, I don't have a preference about sorting them or using the order the author provided in `metadata.yml`.

Given that, I actually would be happy with the output in cell 94 of your notebook above.  It's a concern if that behavior contradicts the pandas documentation though.

> do you think affiliations should be a list in metadata.yaml?

Would the syntax be `affiliations: ['affiliation A', 'affiliation B']` ?

I think that is more confusing to a non-programmer than `affiliations: affiliation A; affiliation B`.  If we eventually want to reach a broad audience, I'd suggest the simplest syntax possible as the default even if the list is more faithful to the YAML specification.

I'll review once we finalize the preferred numbering behavior.

dhimmel:
> I think that the affiliation superscripts should always be sorted

Cool. I ensure that they're sorted, so their sorted order isn't dependent on the current pandas bug.

> If an author introduces multiple new affiliations, I don't have a preference about sorting them or using the order the author provided in metadata.yml.

Currently, it uses the author provided order, which I think makes sense.

> Would the syntax be `affiliations: ['affiliation A', 'affiliation B']` ?

Yeah that's valid YAML for a list I believe. It could also be like:

```yaml
affiliations:
  - affiliation A
  - affiliation B
```

Given your preference for the impure method and its backwards compatibility, we'll keep supporting it.

agitter:
```yaml
affiliations:
  - affiliation A
  - affiliation B
```
seems user-friendly to me.  It avoids the Python list syntax.

Should we use that as the default (i.e. the example we give in manubot-rootstock)?  Then I'd be comfortable dropping backwards compatibility based on my projects.  It would be easy to update deep review.  I'm not sure if any other manubot users are depending on semicolon-delimited affiliations though.

dhimmel:
> seems user-friendly to me. It avoids the Python list syntax.

Our `metadata.yaml` template [already uses](https://github.com/greenelab/manubot-rootstock/blob/f165f609f33b11fdf71a0db6435d4dd159f23973/content/metadata.yaml#L4-L6) this non-pythonic but instead yamlic syntax in other places, [such as](https://github.com/greenelab/manubot-rootstock/blob/f165f609f33b11fdf71a0db6435d4dd159f23973/content/metadata.yaml#L4-L6):

```yaml
keywords:
  - markdown
  - publishing
  - manubot
```

I think the main issue to decide is whether the additional complexity to support both methods is worse than making a backwards incompatible change.

Supporting both methods has the benefit of making the program more flexible. It will be less likely that users run into errors, since the parser is more accommodating. However, it can also be confusing to users when there are two ways to do the same thing. Also there's potentially an added documentation burden. As well as greater technical debt.

agitter:
I'm not familiar with best practices for non-backwards compatible changes in software development, but I would be comfortable making a breaking change at this point.  There isn't an initial manubot release yet, so I expect users understand that behavior can change.

Supporting both lists and semicolon delimiters does seem like it could cause errors in the future, so I am fine moving to lists only.  We can easily update meta review and deep review ourselves.  Most other manubot users with GitHub repos don't have multiple affiliations.  @vsmalladi is an [exception](https://github.com/vsmalladi/tfsee-manuscript/blob/4bbb4ff5c9f12c9e0b8c70f4ee5f23bb16d36e10/content/metadata.yaml) so I'm notifying him about this potential behavior change.

dhimmel:
[Relevant tests](https://travis-ci.org/greenelab/manubot/builds/326900579#L1176) are passing

> best practices for non-backwards compatible changes in software development

I created a `DeprecationWarning`, so if you're not using a list, the program should work, but you will get warned to switch to a list. This way users are given some time to fix the issue. Perhaps overkill in this case, but will be good practice for future breaking changes.

manubot/manubot.py
```diff
@@ -194,6 +195,12 @@ def add_author_affiliations(variables):
         if 'affiliations' not in author:
             continue
         if not isinstance(author['affiliations'], list):
+            warnings.warn(
```
agitter: Will this warning appear if an author has a single affiliation?

tests/manuscripts/example/content/metadata.yaml
```diff
@@ -9,13 +9,16 @@ author_info:
...
     funders: GBMF4552
   -
     github: agitter
     name: Anthony Gitter
     initials: AG
     orcid: 0000-0002-5324-9833
     twitter: anthonygitter
-    affiliations: Department of Biostatistics and Medical Informatics, University of Wisconsin-Madison and Morgridge Institute for Research
+    affiliations:
+      - Department of Biostatistics and Medical Informatics, University of Wisconsin-Madison and Morgridge Institute for Research
```
agitter: We can split my affiliations as well to avoid confusion with the `and`:
```yaml
affiliations:
       - Department of Biostatistics and Medical Informatics, University of Wisconsin-Madison
       - Morgridge Institute for Research
```

manubot/manubot.py
dhimmel: If there affiliations are a list with one affiliation it will not occur.

agitter: Thanks.  I wasn't sure how the YAML would be parsed and whether the affiliation would be a string or a single element list.

PR #24, opened 2018-01-09
PubMed ID citation support

dhimmel:
This pull request migrates the previous `pmid` citations functionality (from NCBI Citation Exporter) to `pmcid`. This is more accurate because the NCBI Citation Exporter only works for records in PubMed Central. `pmid` citations are now supported that use the NCBI E-Utilities. `pubmed.py` enables parsing the XML returned by from E-Utilities into JSON CSL metadata.

Closes https://github.com/greenelab/manubot/issues/21
Refs https://github.com/ncbi/citation-exporter/issues/3

### DIFF
--- a/manubot/pubmed.py
+++ b/manubot/pubmed.py
@@ -22,12 +22,17 @@ def get_pubmed_citeproc(pmid):
     response = requests.get(url, params)
     try:
         element_tree = xml.etree.ElementTree.fromstring(response.text)
-        element_tree, = element_tree.findall('PubmedArticle')
+        element_tree, = list(element_tree)
     except Exception as error:
         logging.error(f'Error fetching PubMed metadata for {pmid}.\n'
-                      f'Invalid response from {response.url}:\n{response.text}')
+                      f'Invalid XML response from {response.url}:\n{response.text}')
         raise error
-    citeproc = citeproc_from_pubmed_article(element_tree)
+    try:
+        citeproc = citeproc_from_pubmed_article(element_tree)
+    except Exception as error:
+        msg = f'Error parsing the following PubMed metadata for PMID {pmid}:\n{response.text}'
+        logging.error(msg)
+        raise error
     return citeproc
 
 
@@ -38,6 +43,9 @@ def citeproc_from_pubmed_article(article):
     https://github.com/citation-style-language/schema/blob/master/csl-data.json
     """
     citeproc = collections.OrderedDict()
+
+    if not article.find("MedlineCitation/Article"):
+        raise NotImplementedError('Unsupported PubMed record: no <Article> element')
 
     title = article.findtext("MedlineCitation/Article/ArticleTitle")
     if title:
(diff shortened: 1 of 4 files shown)


tests/manuscripts/example/content/02.body.md
```diff
@@ -8,3 +8,5 @@ The website Sci-Hub, now in its fifth year of existence, provides gratis access
 Sci-Hub brands itself as "the first pirate website in the world to provide mass and public access to tens of millions of research papers."
 The website, started in 2011, is run by Alexandra Elbakyan, a graduate student and native of Kazakhstan who now resides in Russia [@doi:10.1126/science.aaf5675; @doi:10.1038/nature.2015.18876].
 Elbakyan describes herself as motivated to provide universal access to knowledge [@url:https://engineuring.wordpress.com/2016/03/11/sci-hub-is-a-goal-changing-the-system-is-a-method/; @url:https://www.courtlistener.com/docket/4355308/50/elsevier-inc-v-sci-hub/; @url:http://www.leafscience.org/alexandra-elbakyan/].
+
+Cite the same paper in three ways [@pmid:25648772; @pmcid:PMC4304851; @doi:10.7717/peerj.705].
```
agitter: Is the behavior that each is treated as a separate reference?  We may want to note that behavior in the manubot-rootstock usage instructions.

manubot/pubmed.py
```diff
@@ -0,0 +1,151 @@
...
+        value = article.findtext(xpath)
+        if value:
+            citeproc[key] = value.lower() if key == 'DOI' else value
+
+    abstract = article.findtext("MedlineCitation/Article/Abstract/AbstractText")
+    if abstract:
+        citeproc['abstract'] = abstract
+
+    citeproc['URL'] = f"https://www.ncbi.nlm.nih.gov/pubmed/{citeproc['PMID']}"
+    citeproc['type'] = 'article-journal'
```
agitter: Is `article-journal` always the article type?  PubMed apparently has [books](https://www.nlm.nih.gov/bsd/disted/pubmedtutorial/010_070.html) as well.  As an example, https://www.ncbi.nlm.nih.gov/pubmed/29227604.  It looks like a few other things like the `title` (`BookTitle` in the XML) will fail for books.

I'm okay leaving this as an uncovered corner case, but we may want to document that.

manubot/pubmed.py
```diff
@@ -0,0 +1,151 @@
...
+    'Jul': 7,
+    'Aug': 8,
+    'Sep': 9,
+    'Oct': 10,
+    'Nov': 11,
+    'Dec': 12,
+}
+
+
+def extract_publication_date_parts(article):
```
agitter: Possibly repeat the documentation that `article is a PubmedArticle xml element tree`

manubot/pubmed.py
dhimmel: Good catch. Previously this would fail. But now I made the failure better documented and have it throw a `NotImplementedError`. Added a test case

tests/manuscripts/example/content/02.body.md
dhimmel: Yeah

PR #38, opened 2018-08-03
Configure manubot upload to PyPI

dhimmel:
Figure we should do this while https://pypi.org/project/manubot/ is still available.


agitter:
Great idea.  I don't have experience using Travis to deploy to PyPI so I don't have technical comments on the pull request.  It looks consistent with https://docs.travis-ci.com/user/deployment/pypi/

Now that you're tagging versions, do you also want to enable Zenodo archiving?

dhimmel:
> Now that you're tagging versions, do you also want to enable Zenodo archiving?

Let's hold off on enabling Zenodo archiving for a bit. I'd also like to check out Figshare's alternative repo archiving infrastructure. My concern with these services however is that they require maintainance (manual edits whenever we make a release). IIRC when using Zenodo, sometimes you must update the record title. Also authorship is difficult to determine (we'd need a policy).

I will go ahead and merge and try to get manubot uploaded to PyPI. May take a couple commits if things don't go as planned.

PR #41, opened 2018-08-04
Move citation code into manubot/cite

dhimmel:
WIP


agitter:
> Yeah probably 0.2.0 should be the next version number.

Makes sense to me.

> Somewhat AFAIK.

Thanks.  I'm not that familiar with package conventions.

PR #43, opened 2018-08-05
macOS builds via Travis CI

dhimmel:
Also removes Pandoc installation from CI, since it's not actually needed at the moment (may be in future).


dhimmel:
Wow Travis doesn't have a Python environment for macOS as per https://github.com/travis-ci/travis-ci/issues/2312. So we'd have to install python ourselves.

PR #49, opened 2018-08-07
Prune CSL Items to validate JSON schema

dhimmel:
Closes https://github.com/greenelab/manubot/issues/47

Work in progress. Needs testing and consideration of alternative implementations.

### DIFF
--- a/manubot/cite/citeproc.py
+++ b/manubot/cite/citeproc.py
@@ -1,42 +1,8 @@
-# Valid CSL (citeproc JSON) types as per
-# https://github.com/citation-style-language/schema/blob/4846e02f0a775a8272819204379a4f8d7f45c16c/csl-types.rnc#L5-L39
-citeproc_types = {
-    "article",
-    "article-journal",
-    "article-magazine",
-    "article-newspaper",
-    "bill",
-    "book",
-    "broadcast",
-    "chapter",
-    "dataset",
-    "entry",
-    "entry-dictionary",
-    "entry-encyclopedia",
-    "figure",
-    "graphic",
-    "interview",
-    "legal_case",
-    "legislation",
-    "manuscript",
-    "map",
-    "motion_picture",
-    "musical_score",
-    "pamphlet",
-    "paper-conference",
-    "patent",
-    "personal_communication",
-    "post",
-    "post-weblog",
-    "report",
-    "review",
-    "review-book",
-    "song",
-    "speech",
-    "thesis",
-    "treaty",
-    "webpage",
-}
+import copy
+import functools
+
+import jsonref
+import jsonschema
 
 citeproc_type_fixer = {
     'journal-article': 'article-journal',
@@ -47,26 +13,12 @@ citeproc_type_fixer = {
     'reference-entry': 'entry',
 }
 
-# Remove citeproc keys to fix pandoc-citeproc errors
-citeproc_remove_keys = [
-    # Error in $[0].ISSN[0]: failed to parse field ISSN: mempty
-    'ISSN',
-    # Error in $[2].ISBN[0]: failed to parse field ISBN: mempty
-    'ISBN',
-    # pandoc-citeproc expected Object not array for archive
-    'archive',
-    # failed to parse field event: Could not read as string
-    'event',
-    # remove the references of cited papers. Not neccessary and unwieldy.
-    'reference',
-    # Error in $[26].categories[0][0]: failed to parse field categories: mempty
-    'categories',
-]
-
 
 def citeproc_passthrough(csl_item, set_id=None):
     """
-    Fix errors in a CSL item and optionally change its id.
+    Fix errors in a CSL item, according to the CSL JSON schema, and optionally
+    change its id.
+
     http://citeproc-js.readthedocs.io/en/latest/csl-json/markup.html
     https://github.com/citation-style-language/schema/blob/master/csl-data.json
     """
@@ -75,23 +27,66 @@ def citeproc_passthrough(csl_item, set_id=None):
 
     # Correct invalid CSL item types
     # See https://github.com/CrossRef/rest-api-doc/issues/187
-    old_type = csl_item['type']
-    csl_type = citeproc_type_fixer.get(old_type, old_type)
-    if csl_type not in citeproc_types:
-        csl_type = 'entry'
-    csl_item['type'] = csl_type
+    csl_item['type'] = citeproc_type_fixer.get(csl_item['type'], csl_item['type'])
 
-    # Remove problematic objects
-    for key in citeproc_remove_keys:
-        csl_item.pop(key, None)
+    # Remove fields that violate the CSL Item JSON Schema
+    validator = get_jsonschema_csl_validator()
+    csl = [csl_item]
+    errors = list(validator.iter_errors(csl))
+    csl_item, = remove_jsonschema_errors(csl, errors)
 
-    # pandoc-citeproc error
-    # failed to parse field issued: Could not read as string: Null
-    try:
-        value = csl_item['issued']['date-parts'][0][0]
-        if value is None:
-            del csl_item['issued']
-    except KeyError:
-        pass
+    # Default CSL type to entry
+    csl_item['type'] = csl_item.get('type', 'entry')
 
+    # Confirm that corrected CSL validates
+    validator.validate([csl_item])
     return csl_item
+
+
+@functools.lru_cache()
+def get_jsonschema_csl_validator():
+    """
+    Return a jsonschema validator for the CSL Item JSON Schema
+    """
+    url = 'https://github.com/citation-style-language/schema/raw/4846e02f0a775a8272819204379a4f8d7f45c16c/csl-data.json'
+    # Use jsonref to workaround https://github.com/Julian/jsonschema/issues/447
+    schema = jsonref.load_uri(url, jsonschema=True)
+    # Cannot yet infer draft from schema https://github.com/citation-style-language/schema/pull/153
+    jsonschema.Draft3Validator.check_schema(schema)
+    return jsonschema.Draft3Validator(schema)
+
+
+def remove_jsonschema_errors(instance, errors):
+    """
+    Remove fields that produced JSON Schema errors.
+    https://github.com/Julian/jsonschema/issues/448
+    """
+    instance = copy.deepcopy(instance)
+    errors = sorted(errors, key=lambda e: e.path, reverse=True)
+    for error in errors:
+        _remove_error(instance, error)
+    return instance
+
+
+def _delete_elem(instance, path):
+    """
+    Helper function for remove_jsonschema_errors
+    """
+    *head, tail = path
+    for key in head:
+        instance = instance[key]
+    del instance[tail]
+
+
+def _remove_error(instance, error):
+    """
+    Helper function for remove_jsonschema_errors
+    """
+    if error.validator == 'additionalProperties':
+        extras = set(error.instance) - set(error.schema['properties'])
+        for key in extras:
+            _delete_elem(instance, path=list(error.path) + [key])
+    elif error.validator in {'enum', 'type'}:
+        _delete_elem(instance, error.path)
+    else:
+        raise NotImplementedError(f'{error.validator} is not yet supported')
--- a/setup.py
+++ b/setup.py
@@ -50,6 +50,8 @@ setuptools.setup(
     install_requires=[
         'errorhandler',
         'jinja2',
+        'jsonref',
+        'jsonschema',
         'pandas',
         'pybase62',
         'pyyaml',


dhimmel:
As of https://github.com/greenelab/manubot/pull/49/commits/ef7adebff47630b5d6221b865bcfaf5e4ee002db, the REAMDE command returns:

```
manubot cite doi:10.1098/rsif.2017.0387 pmid:29424689 pmcid:PMC5640425 arxiv:1806.05726
```

<details>

```json
[
  {
    "publisher": "The Royal Society",
    "issue": "141",
    "DOI": "10.1098/rsif.2017.0387",
    "type": "article-journal",
    "page": "20170387",
    "source": "Crossref",
    "title": "Opportunities and obstacles for deep learning in biology and medicine",
    "volume": "15",
    "author": [],
    "container-title": "Journal of The Royal Society Interface",
    "language": "en",
    "issued": {
      "date-parts": [
        [
          2018,
          4
        ]
      ]
    },
    "URL": "https://doi.org/10.1098/rsif.2017.0387",
    "container-title-short": "J. R. Soc. Interface",
    "PMCID": "PMC5938574",
    "PMID": "29618526",
    "id": "PZMP42Ak"
  },
  {
    "title": "Sci-Hub provides access to nearly all scholarly literature.",
    "volume": "7",
    "container-title": "eLife",
    "container-title-short": "Elife",
    "ISSN": "2050-084X",
    "issued": {
      "date-parts": [
        [
          2018,
          3,
          1
        ]
      ]
    },
    "author": [
      {
        "given": "Daniel S",
        "family": "Himmelstein"
      },
      {
        "given": "Ariel Rodriguez",
        "family": "Romero"
      },
      {
        "given": "Jacob G",
        "family": "Levernier"
      },
      {
        "given": "Thomas Anthony",
        "family": "Munro"
      },
      {
        "given": "Stephen Reid",
        "family": "McLaughlin"
      },
      {
        "given": "Bastian",
        "family": "Greshake Tzovaras"
      },
      {
        "given": "Casey S",
        "family": "Greene"
      }
    ],
    "PMID": "29424689",
    "PMCID": "PMC5832410",
    "DOI": "10.7554/elife.32822",
    "abstract": "The website Sci-Hub enables users to download PDF versions of scholarly articles, including many articles that are paywalled at their journal's site. Sci-Hub has grown rapidly since its creation in 2011, but the extent of its coverage has been unclear. Here we report that, as of March 2017, Sci-Hub's database contains 68.9% of the 81.6 million scholarly articles registered with Crossref and 85.1% of articles published in toll access journals. We find that coverage varies by discipline and publisher, and that Sci-Hub preferentially covers popular, paywalled content. For toll access articles, we find that Sci-Hub provides greater coverage than the University of Pennsylvania, a major research university in the United States. Green open access to toll access articles via licit services, on the other hand, remains quite limited. Our interactive browser at <ext-link ext-link-type=\"uri\" xlink:href=\"https://greenelab.github.io/scihub\">https://greenelab.github.io/scihub</ext-link> allows users to explore these findings in more detail. For the first time, nearly all scholarly literature is available gratis to anyone with an Internet connection, suggesting the toll access business model may become unsustainable.",
    "URL": "https://www.ncbi.nlm.nih.gov/pubmed/29424689",
    "type": "article-journal",
    "id": "IJbk7Kpe"
  },
  {
    "source": "PMC",
    "accessed": {
      "date-parts": [
        [
          2018,
          8,
          7
        ]
      ]
    },
    "id": "HRLhuqR3",
    "title": "Systematic integration of biomedical knowledge prioritizes drugs for repurposing",
    "author": [
      {
        "family": "Himmelstein",
        "given": "Daniel Scott"
      },
      {
        "family": "Lizee",
        "given": "Antoine"
      },
      {
        "family": "Hessler",
        "given": "Christine"
      },
      {
        "family": "Brueggeman",
        "given": "Leo"
      },
      {
        "family": "Chen",
        "given": "Sabrina L"
      },
      {
        "family": "Hadley",
        "given": "Dexter"
      },
      {
        "family": "Green",
        "given": "Ari"
      },
      {
        "family": "Khankhanian",
        "given": "Pouya"
      },
      {
        "family": "Baranzini",
        "given": "Sergio E"
      }
    ],
    "editor": [
      {
        "family": "Valencia",
        "given": "Alfonso"
      }
    ],
    "container-title-short": "eLife",
    "container-title": "eLife",
    "publisher": "eLife Sciences Publications, Ltd",
    "ISSN": "2050-084X",
    "issued": {
      "date-parts": [
        [
          2017
        ]
      ]
    },
    "page": "e26726",
    "volume": "6",
    "PMID": "28936969",
    "PMCID": "PMC5640425",
    "DOI": "10.7554/eLife.26726",
    "type": "article-journal",
    "URL": "https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5640425/"
  },
  {
    "version": "1",
    "URL": "https://arxiv.org/abs/1806.05726v1",
    "title": "Generalization of the Fermi Pseudopotential",
    "issued": {
      "date-parts": [
        [
          "2018",
          "06",
          "14"
        ]
      ]
    },
    "author": [
      {
        "literal": "Trang T. Le"
      },
      {
        "literal": "Zach Osman"
      },
      {
        "literal": "D. K. Watson"
      },
      {
        "literal": "Martin Dunn"
      },
      {
        "literal": "B. A. McKinney"
      }
    ],
    "container-title": "arXiv",
    "publisher": "arXiv",
    "abstract": "Introduced eighty years ago, the Fermi pseudopotential has been a powerful\nconcept in multiple fields of physics. It replaces the detailed shape of a\npotential by a delta-function operator multiplied by a parameter giving the\nstrength of the potential. For Cartesian dimensions $d>1$, a regularization\noperator is necessary to remove singularities in the wave function. In this\nstudy, we develop a Fermi pseudopotential generalized to $d$ dimensions\n(including non-integer) and to non-zero wavenumber, $k$. Our approach has the\nadvantage of circumventing singularities that occur in the wave function at\ncertain integer values of $d$ while being valid arbitrarily close to integer\n$d$. In the limit of integer dimension, we show that our generalized\npseudopotential is equivalent to previously derived $s$-wave pseudopotentials.\nOur pseudopotential generalizes the operator to non-integer dimension, includes\nenergy ($k$) dependence, and simplifies the dimension-dependent coupling\nconstant expression derived from a Green's function approach. We apply this\npseudopotential to the problem of two cold atoms ($k\\to0$) in a harmonic trap\nand extend the energy expression to arbitrary dimension.",
    "type": "report",
    "id": "10GH4uYUR"
  }
]
```

</details>

Much simpler.

dhimmel:
Opened https://github.com/Julian/jsonschema/issues/449 regarding testing failure:

```
E   TypeError: attrib() got an unexpected keyword argument 'converter'
```

manubot/cite/citeproc.py
```diff
@@ -47,51 +14,133 @@
...
+        instance = instance[key]
+    return instance
+
+
+def _remove_error(instance, error):
+    """
+    Remove a jsonschema ValidationError from the JSON-like instance.
+
+    See ValidationError documentation at
+    http://python-jsonschema.readthedocs.io/en/latest/errors/#jsonschema.exceptions.ValidationError
```
agitter: It's somewhat difficult to follow this logic without reading the `ValidationError` documentation.  It's probably okay because you link the documentation for any future maintainers.

manubot/cite/doi.py
```diff
@@ -42,7 +42,7 @@ def get_doi_citeproc(doi):
     citeproc['URL'] = f'https://doi.org/{doi}'
     short_doi_url = get_short_doi_url(doi)
     if short_doi_url:
-        citeproc['short_url'] = short_doi_url
+        citeproc['URL'] = short_doi_url
```
agitter: If I follow correctly, this will be the only URL that is displayed per the Manubot CSL style.  I prefer the longer URL for DOIs because the DOI can be easily extracted from the reference list.

tests/csl-json/crossref-deep-review-csl/raw.json
```diff
@@ -0,0 +1,3814 @@
...
+    "published-online": {
+      "date-parts": [
+        [
+          2018,
+          4,
+          4
+        ]
+      ]
+    },
+    "reference": [
```
agitter: Wow, I didn't realize all the references were included in the raw JSON.  This is a great case for pruning.

tests/csl-json/test_csl_pruning.py
```diff
@@ -0,0 +1,20 @@
...
+from manubot.cite.citeproc import remove_jsonschema_errors
+
+directory = pathlib.Path(__file__).parent
+csl_instances = [
+    x.name for x in directory.glob('*-csl')
+]
+
+
+@pytest.mark.parametrize('name', csl_instances)
+def test_remove_jsonschema_errors(name):
```
agitter: These tests cover most of the types of CSL JSON files Manubot would receive from its typical sources.  They don't tell us how pruning will behave one other sources.  That should be okay, but I'm thinking about whether we want to add a comment for maintenance.  If Manubot adds new metadata sources, those should probably have their pruning tested here as well.

Do we need to test pruning a URL reference?

README.md
```diff
@@ -108,6 +108,8 @@ positional arguments:
 optional arguments:
   -h, --help   show this help message and exit
   --file FILE  specify a file to write CSL output, otherwise default to stdout
+  --bad-csl    allow CSL Items that do not conform to the JSON Schema. Skips
```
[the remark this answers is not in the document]
dhimmel: I think `--bad-csl` or `--allow-bad-csl` is more accurate than "raw". There is a schema for what CSL should be and Crossref DOI content negotiation returns fields and values that are not valid CSL. In the past, we were patching specific invalid CSL fields that were causing pandoc-citeproc to fail. However, we were always playing catch up (an error would occur and we would debug it). By pruning, we're removing invalid (i.e. bad) CSL up front so it doesn't cause downstream issues or clutter `references.json`.

manubot/cite/arxiv.py
```diff
@@ -46,7 +46,7 @@ def get_arxiv_citeproc(arxiv_id):
     pattern = re.compile(r'arxiv.org/abs/(.+)')
     match = pattern.search(url)
     versioned_id = match.group(1)
-    csl_item['arxiv_id'] = versioned_id
+    csl_item['archive_location'] = versioned_id
```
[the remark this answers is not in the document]
dhimmel: Yeah I also had trouble deciding what field to use.

Source is defined as:

> from whence the item originates (e.g. a library catalog or database)

So the source here would probably be `arXiv`, as opposed to `PubMed` or `Crossref`.

Number [can be](https://github.com/citation-style-language/schema/blob/04504e8a1ec53056e2b540c533b8bdbea805aded/csl-data.json#L363-L368) a string or a number, so we could actually do this.

However, given that archive_location refers to a physical location (which is less relevant), I though it was ripe for repurposing. Basically, I though the arxiv id is the digital equivalent of a location in an archive.

manubot/cite/citeproc.py
```diff
@@ -47,51 +14,133 @@
...
     # Correct invalid CSL item types
     # See https://github.com/CrossRef/rest-api-doc/issues/187
-    old_type = csl_item['type']
-    csl_type = citeproc_type_fixer.get(old_type, old_type)
-    if csl_type not in citeproc_types:
-        csl_type = 'entry'
-    csl_item['type'] = csl_type
-
-    # Remove problematic objects
-    for key in citeproc_remove_keys:
```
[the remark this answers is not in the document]
dhimmel: If prune=False, problematic keys/values are not removed. Oftentimes the invalid CSL is harmless, but sometimes is causes pandoc-citeproc to break.

> Is the assumption that whenever Manubot is used with pandoc the default prune=True will be used?

I was planning to eventually refactor the code such that `--bad-csl` was also an option for `manubot process`, although it should be highly discouraged.

manubot/cite/doi.py
dhimmel: Yes that is true, but I was also hoping to update the default style so a reference renders differently.

Currently:

33. **Sci-Hub provides access to nearly all scholarly literature**
Daniel S Himmelstein, Ariel Rodriguez Romero, Jacob G Levernier, Thomas Anthony Munro, Stephen Reid McLaughlin, Bastian Greshake Tzovaras, Casey S Greene
_eLife_ (2018-03-01) https://doi.org/10.7554/elife.32822

Envisioned:

34. **Sci-Hub provides access to nearly all scholarly literature**
Daniel S Himmelstein, Ariel Rodriguez Romero, Jacob G Levernier, Thomas Anthony Munro, Stephen Reid McLaughlin, Bastian Greshake Tzovaras, Casey S Greene
_eLife_ (2018-03-01) https://doi.org/ckcj
DOI: 10.7554/elife.32822 · PMID: 29424689 · PMCID: PMC5832410

Now possible due to https://github.com/greenelab/manubot/issues/45. This would make a more uniform reference across DOI and PMID / PMCID citations.

A style could always decide to recreate the URL from a DOI if available, if they wanted full DOI links. @agitter what do you think?

tests/csl-json/crossref-deep-review-csl/raw.json
dhimmel: Yes Crossref started including them. Note we already were filtering them from `references.json` as per:

https://github.com/greenelab/manubot/blob/fdafa3eb5969662598e5a602d14e161568a59a83/manubot/cite/citeproc.py#L50-L64

This PR generalizes the pruning, such that we don't need `citeproc_remove_keys`.

tests/csl-json/test_csl_pruning.py
dhimmel: > Do we need to test pruning a URL reference?

Yes.

>  If Manubot adds new metadata sources, those should probably have their pruning tested here as well.

Definitely! I can see where the best place to document this is.

dhimmel: Note it's not as simple as one test per metadata source. For example, DOI content negotiation returns different raw CSL based on the registrar. Also we're taking the output at a single point in time. These metadata endpoints change their responses over time. For example, previously Crossref didn't list all the references. Now it does.

dhimmel:
AppVeyor builds failing due to what seems like a network connectivity issue:

> [NuGet] Could not connect to the feed specified at 'https://chocolatey.org/api/v2/'. Please verify that the package source (located in the Package Manager Settings) is valid and ensure your network connectivity.

manubot/cite/citeproc.py
dhimmel: Yeah, I myself don't even have a great understanding of the `jsonschema` package. It's possible there are types of errors that could occur that this function doesn't handle. However, with the current CSL Schema, I'm not sure they will come up.

dhimmel: In the case of this `_remove_error` function, the fewer comments the better as there is a chance my comments would be wrong. In other words, to understand it, one should reference the code. I agree it's unfortunately complex.

dhimmel:
@agitter ready for re-review

README.md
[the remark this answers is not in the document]
agitter: How about changing it to `--allow-bad-csl`?  `--bad-csl` gave me the impression (before reading the help message and code) that manubot is doing something bad or invalid.

manubot/cite/arxiv.py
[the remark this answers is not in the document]
agitter: `archive_location` is okay with me if that is still your preference.

`number` seems more intuitive in terms of matching existing CSL style files.

`source` and `note` aren't very good options.

manubot/cite/citeproc.py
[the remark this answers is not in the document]
agitter: Okay, this behavior makes sense.  We don't want to support pruning and the ad hoc corrections.

manubot/cite/doi.py
agitter: > A style could always decide to recreate the URL from a DOI if available, if they wanted full DOI links.

Is that possible using the CSL style alone?  If it is, then switching to the short DOI URLs and new default style sounds good to me.  If it would require editing the CSL JSON or some other processing outside the style file, then I'm not so sure.

tests/csl-json/test_csl_pruning.py
agitter: The new test and docstring are very helpful.

manubot/cite/arxiv.py
```diff
@@ -56,7 +56,7 @@ def get_arxiv_citeproc(arxiv_id):
     # Extract CSL date field
     published = entry.findtext(prefix + 'published')
     published, _ = published.split('T', 1)
-    csl_item['issued'] = {'date-parts': [published.split('-')]}
+    csl_item['issued'] = {'date-parts': [[int(x) for x in published.split('-')]]}
```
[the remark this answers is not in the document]
dhimmel: It's not necessary as per the schema which allows int or number, but our other citeproc functions return ints, so I thought it'd be good for standardization.

manubot/cite/doi.py
dhimmel: > Is that possible using the CSL style alone?

Yes.

README.md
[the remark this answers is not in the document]
dhimmel: Now I'm starting to favor the more verbose `--allow-invalid-csl-data`. We don't necessarily expect people to be using this argument very much?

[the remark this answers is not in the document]
dhimmel: Done in https://github.com/greenelab/manubot/pull/49/commits/8ca1a44cfd101b9d8fb99be7966ae0db021e4df3

dhimmel:
@agitter  ready for re-review again.

README.md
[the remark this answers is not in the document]
agitter: `--allow-invalid-csl-data` is good and fits the behavior well.  The more compact `--allow-invalid-csl`  would also be okay and has the same meaning in my opinion.

manubot/cite/doi.py
agitter: I still have a slight preference for the longer URL, but I'm okay switching to the short URL given that it can be adjusted from the style file..

README.md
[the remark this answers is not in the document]
dhimmel: Yes, both are OKAY, but I'm trying to be better about differentiating between CSL (which usually refers to the XML style) and CSL Data which refers to the JSON references. Note also that Python's argparse will understand `--allow-invalid-csl` or even `--allo` as being this argument (although we don't necessarily need to advertise this).

I will keep as `--allow-invalid-csl-data`, but we should revisit the arguments after the next couple PRs before the next release to make sure we're happy.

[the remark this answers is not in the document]
agitter: > Note also that Python's argparse will understand --allow-invalid-csl or even --allo

Neat, I didn't know that.  The distinction between CSL style and data is confusing.  I see your point there.

PR #50, opened 2018-08-12
Test fidelity of README codeblocks with command ouput

dhimmel:
~~Rebase on #49 once merged.~~ This PR is now independent of #49.

### DIFF
--- a/tests/test_readme.py
+++ b/tests/test_readme.py
@@ -1,0 +1,38 @@
+import pathlib
+import re
+import subprocess
+
+import pytest
+
+readme_path = pathlib.Path(__file__).parent.parent / 'README.md'
+readme = readme_path.read_text()
+
+pattern = r'''\
+<!-- test codeblock contains output of `(?P<command>.+?)` -->
+```
+(?P<output>.+?)
+```
+'''
+pattern = re.compile(pattern, re.DOTALL)
+matches = list(pattern.finditer(readme))
+matches
+
+
+@pytest.mark.parametrize(
+    ['command', 'expected'],
+    [match.groups() for match in matches],
+)
+def test_readme_codeblock_contains_output_from(command, expected):
+    """
+    If this test fails, ensure that codeblocks in README.md have the correct
+    output. To enable this check for output in a codeblock, use the following
+    construct:
+
+    <!-- test codeblock contains output of `{command}` -->
+    ```
+    {expected}
+    ```
+    """
+    output = subprocess.check_output(command, shell=True)
+    output = output.decode().rstrip()
+    assert output == expected
--- a/tests/test_readme.py
+++ b/tests/test_readme.py
@@ -33,6 +33,5 @@ def test_readme_codeblock_contains_output_from(command, expected):
     {expected}
     ```
     """
-    output = subprocess.check_output(command, shell=True)
-    output = output.decode().rstrip()
-    assert output == expected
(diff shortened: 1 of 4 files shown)


dhimmel:
I've requested a review from @ajlee21. In the Greene Lab we have a code review policy, so since @agitter may be busy I thought it may make sense to use an internal review for this PR.

tests/test_readme.py
```diff
@@ -0,0 +1,59 @@
...
+    argnames=['command', 'expected'],
+    argvalues=[match.groups() for match in matches],
+    ids=[match.group('command') for match in matches],
+)
+def test_readme_codeblock_contains_output_from(command, expected):
+    """
+    If this test fails, ensure that codeblocks in README.md have the correct
+    output. To enable this check for output in a codeblock, use the following
+    construct:
+
```
ajlee21: Not necessary for merging, but I'm not sure what you mean by "codeblocks" here

tests/test_readme.py
```diff
@@ -0,0 +1,59 @@
...
+    ```
+    {expected}
+    ```
+    """
+    output = _get_output_from(command)
+    assert output == expected
+
+
+def _get_output_from(command):
+    return subprocess.check_output(command, shell=True, universal_newlines=True)
```
ajlee21: Maybe add some comments to describe what each function is testing for

tests/test_readme.py
dhimmel: These things (https://help.github.com/articles/creating-and-highlighting-code-blocks/) in the markdown README for this repo.

tests/test_readme.py
dhimmel: This is just a helper function. The tests all begin with `test_` (pytest is smart like that)!

This function is just a wrapper around subprocess.check_output, so we don't have to repeat the command / arguments twice in two parts of this file.

Will take a look to make sure all tests have docstrings that this PR adds.

tests/test_readme.py
```diff
@@ -0,0 +1,59 @@
+import pathlib
+import re
+import subprocess
+
+import pytest
+
+readme_path = pathlib.Path(__file__).parent.parent / 'README.md'
```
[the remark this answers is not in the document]
dhimmel: No none of the tests will, since the `tests` directory does not get bundled into our python package. `pathlib.Path(__file__)` will work when packaged, although .parent.parent won't navigate to a directory with `README.md`.

tests/test_readme.py
ajlee21: I see, thanks.  I think I'm not exactly sure what this project is/what the goal of this project is so perhaps my comments don't make too much sense

dhimmel: My bad! I should have given you a bit more context. This repository creates the `manubot` command line program written in Python. The README contains some codeblocks which are supposed to contain the output produced by running a specific manubot command, such as `manubot --help`. This PR implements a test to check that the output in the README codeblocks is what it should be given the current manubot program. Basically, it's helpful to remind someone they forgot to update the README. It also implements a script to update the README for you.

tests/test_readme.py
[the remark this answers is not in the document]
dhimmel: >  seems safe to assume this code would only be run within the manubot repo directory structure

Yes, this should be the assumption for everything in `tests`

PR #54, opened 2018-08-21
Move manubot cite command infrastructure to submodule

dhimmel:


dhimmel:
Ready for review (can't think of anything else I want to do right now). It moves a lot of stuff around with a few minor actual changes.

I may consider the traditional merge commit for this PR to keep the commits distinct.

PR #56, opened 2018-08-22
Enable lazy imports by the CLI & clear __init__.py files

dhimmel:
Enables manubot CLI to be called without running these files and triggering the imports. #53 and #54 also aimed partially to do this but were incomplete.

How I noticed the issue was that when running `manubot --version`, I got the following output:

```
/home/dhimmel/anaconda3/envs/manubot-dev/lib/python3.6/importlib/_bootstrap.py:219: RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88
  return f(*args, **kwds)
v0.1.0
```

Basically, when we did `from manubot.cite.cite_command import add_subparser_cite`, this was executing `manubot.cite.__init__.py`. For `manubot.process.__init__.py`, this was importing pandas and other heavy dependencies, which:

1. cause annoying messages like the one above
2. slow down basic command line calls
3. increase the change of improperly configured environments leading to unnecessary failure.

This PR is one approach to fix this. However, it has the aesthetically unfortunate consequence of changing the API to now need to specify another submodule, e.g. `import manubot.cite.cite`. However, it will also allow importing single functions from say `manubot.cite.doi` without having to run the entire __init__.py file.

An alternative solution would be to move `add_subparser_cite` and `add_subparser_process` to `manubot.command.py` and change the line `parser.set_defaults(function=cli_cite)` to use a string to specify the submodule/function, which would then be [imported at runtime](https://stackoverflow.com/a/8790232/4651668) via importlib.

@agitter what do you think


Miserlou:
The stuff currently in `__init__.py` should moved to a `util.py`. I _think_ you can then put your desired imports in `__init__.py` to avoid the `cite.cite.blah` ugliness.

kurtwheeler:
I don't think you could solve this through organizational changes. While `import manubot.cite.cite` is kinda awkward, I think it may be the best bet. The only thing that I could think of that might  kinda do what you want would be to conditionally import things in your __init__.py files like so:
```
if not IS_SCRIPT:
  from manubot.cite.cite import standardize_citation
  from manubot.cite.cite import is_valid_citation_string
  etc
```
and then set IS_SCRIPT to true from your management commands. I think this is an uglier solution than `import manubot.cite.cite` though.

dhimmel:
Okay, thanks @Miserlou and @kurtwheeler for weighing in. Both comments were helpful:

+ In https://github.com/greenelab/manubot/pull/56/commits/da90fe6a1ac3b466a3b4c2297f38b799f2c2031a, I renamed the former `__init__.py` files to `util.py`. For example, it's now `manubot.cite.util` rather than `manubot.cite.cite`. I think it's a bit less confusing when we don't repeat cite.

+ TODO (will update this comment)

dhimmel:
@agitter do you think we should restore anyimports to `manubot.cite` or `manubot.process`? Does this look good to you?

dhimmel:
In https://github.com/greenelab/manubot/pull/56/commits/4f11fd98480ce807cf60ba198b103005b9cd526e, I exposed `expose citation_to_citeproc` & `standardize_citation` under `manubot.cite`. I reformatted manubot.cite.util to be lazy such that no non-builtin packages are imported unless specific functions are actually called.

PR #76, opened 2018-10-19
Validate citation identifiers conform to known identifier patterns

dhimmel:
Closes https://github.com/greenelab/manubot/issues/71

### DIFF
--- a/manubot/cite/cite_command.py
+++ b/manubot/cite/cite_command.py
@@ -9,6 +9,7 @@ from manubot.cite import (
     citation_to_citeproc,
     standardize_citation,
 )
+from manubot.cite.util import is_valid_citation_string
 
 # For manubot cite, infer --format from --output filename extensions
 extension_to_format = {
@@ -71,8 +72,9 @@ def cli_cite(args):
     # generate CSL JSON data
     csl_list = list()
     for citation in args.citations:
-        citation = standardize_citation(citation)
         try:
+            is_valid_citation_string(f'@{citation}')
+            citation = standardize_citation(citation)
             csl_item = citation_to_citeproc(citation, prune=args.prune_csl)
             csl_list.append(csl_item)
         except Exception as error:
--- a/manubot/cite/cite_command.py
+++ b/manubot/cite/cite_command.py
@@ -73,7 +73,8 @@ def cli_cite(args):
     csl_list = list()
     for citation in args.citations:
         try:
-            is_valid_citation_string(f'@{citation}')
+            if not is_valid_citation_string(f'@{citation}'):
+                continue
             citation = standardize_citation(citation)
             csl_item = citation_to_citeproc(citation, prune=args.prune_csl)
             csl_list.append(csl_item)
--- a/manubot/cite/util.py
+++ b/manubot/cite/util.py
@@ -81,18 +81,22 @@ def inspect_citation_identifier(citation):
     return None
 
 
-def is_valid_citation_string(string, manuscript_extras=False):
+def is_valid_citation_string(
+        string, allow_tag=False, allow_raw=False, allow_pandoc_xnos=False):
     """
     Return True if string is a properly formatted citation. Return False if
     string is not a citation or is an invalid citation.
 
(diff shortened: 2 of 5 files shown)


manubot/cite/util.py
```diff
@@ -42,6 +42,48 @@ def standardize_citation(citation):
...
+def validate_citation(citation):
+    source, identifier = citation.split(':', 1)
+    if source == 'pmid':
+        # https://www.nlm.nih.gov/bsd/mms/medlineelements.html#pmid
+        if identifier.startswith('PMC'):
+            logging.warning(
+                'PubMed Identifiers should start with digits rather than PMC. '
+                f'Should {citation} switch the citation source to `pmcid`?'
+            )
+        if not regexes['pmid'].fullmatch(identifier):
```
dhimmel: change to elif

manubot/cite/util.py
```diff
@@ -42,6 +42,48 @@ def standardize_citation(citation):
...
+                f'Pubmed Central Identifiers must start with `PMC`.'
+            )
+    if source == 'doi':
+        # https://www.crossref.org/blog/dois-and-matching-regular-expressions/
+        if not identifier.startswith('10.'):
+            logging.warning(
+                f'Invalid DOI citation: {citation}\n'
+                'DOIs must start with `10.`.'
+            )
+        if not regexes['doi'].fullmatch(identifier):
```
dhimmel: change to elif

manubot/cite/cite_command.py
```diff
@@ -71,8 +72,10 @@ def cli_cite(args):
     # generate CSL JSON data
     csl_list = list()
     for citation in args.citations:
-        citation = standardize_citation(citation)
         try:
+            if not is_valid_citation_string(f'@{citation}'):
+                continue
```
agitter: Should we log something here, perhaps at the `INFO` level, to report that some of the citation strings were skipped?  I tested
```
$ manubot cite fig:123 eq:56
[]
```

That behavior is valid, but the silent output could be confusing.

manubot/cite/util.py
```diff
@@ -72,6 +120,11 @@ def is_valid_citation_string(string):
         logging.error(f'{string} → source "{source}" is not valid')
```
agitter: We currently require a lower case source.  Could that be confusing to users, and do we want to provide a hint in the error message?  The test
```
$ manubot cite DOI:10.1038/nrg3920
## ERROR
@DOI:10.1038/nrg3920 \u2192 source "DOI" is not valid
```
prompted this.  A novice may not understand why `DOI` doesn't work.

dhimmel:
@agitter great suggestions. I think you will be pleasantly surprised by the last three commits and their ability to address your comments!

manubot/cite/util.py
```diff
@@ -81,19 +81,25 @@ def inspect_citation_identifier(citation):
...
-    string is not a citation (i.e. it's an exempt category such as @fig) or is
-    an invalid citation. In the case string is an invalid citation, an error is
-    logged. This function does not catch all invalid citations, but instead
-    performs cursory checks, such as citations adhere to the expected formats.
-    No calls to external resources are used by these checks, so they will not
-    detect citations to non-existent identifiers unless those identifiers
-    violate their source's syntax.
+    string is not a citation or is an invalid citation.
+
+    In the case string is an invalid citation, an error is logged. This function does not catch all invalid citations, but instead performs cursory checks, such as citations adhere to the expected formats. No calls to external resources are used by these checks, so they will not detect citations to non-existent identifiers unless those identifiers violate their source's syntax.
```
agitter: The line wrapping was lost here.

manubot/cite/util.py
```diff
@@ -106,18 +112,41 @@ def is_valid_citation_string(string):
...
-    # Exempted non-citation sources used for pandoc-fignos, pandoc-tablenos,
-    # and pandoc-eqnos
-    if source in {'fig', 'tbl', 'eq'}:
-        return False
+    if allow_pandoc_xnos:
+        # Exempted non-citation sources used for pandoc-fignos,
+        # pandoc-tablenos, and pandoc-eqnos
+        pandoc_xnos_keys = {'fig', 'tbl', 'eq'}
+        if source in pandoc_xnos_keys:
+            return False
```
agitter: I may be misreading the logic here.  If `allow_pandoc_xnos` and `source in pandoc_xnos_keys`, shouldn't this return True?

agitter: Got it, that behavior makes sense.

PR #77, opened 2018-10-22
Cite command error handling

dhimmel:
Do not exit program when a citation_to_citeproc lookup fails. Instead set a nonzero exit code, but proceed with passing citaitons. Error messages are logged, full stacktrace is logged at INFO level.

### DIFF
--- a/manubot/cite/cite_command.py
+++ b/manubot/cite/cite_command.py
@@ -72,7 +72,12 @@ def cli_cite(args):
     csl_list = list()
     for citation in args.citations:
         citation = standardize_citation(citation)
-        csl_list.append(citation_to_citeproc(citation, prune=args.prune_csl))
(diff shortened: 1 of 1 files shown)


dhimmel:
@slochower I requested a review from you, since I think you know the citation infrastructure somewhat (?) and I was hoping to take some of the reviewer burden off of @agitter.

slochower:
Sure! Will take a look tomorrow.

slochower:
Hmm... I only see one line changed to a `try`/`except`. Is that it? It looks fine; the docs say that `sys.exc_info()` will return the exception information, which looks good.

dhimmel:
Yeah it is a small change, but we try to have every PR get reviewed unless it's extremely urgent.

PR #80, opened 2018-11-14
get_pmc_citeproc: switch to Literature Citation Exporter

dhimmel:
NCBI Citaiton Exporter was taken offline without notice:
https://twitter.com/dhimmel/status/1061787168820092929

### DIFF
--- a/manubot/cite/cite_command.py
+++ b/manubot/cite/cite_command.py
@@ -79,7 +79,10 @@ def cli_cite(args):
             csl_item = citation_to_citeproc(citation, prune=args.prune_csl)
             csl_list.append(csl_item)
         except Exception as error:
-            logging.error(error)
+            logging.error(
+                f'citation_to_citeproc for {citation} failed '
+                f'due to a {error.__class__.__name__}:\n{error}'
+            )
             logging.info(error, exc_info=True)
 
     # output CSL JSON data, if --render is False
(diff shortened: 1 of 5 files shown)


dhimmel:
@agitter this should take precedence over other PRs, since currently any `pmcid` citations will fail.

Currently this PR only switches `get_pmc_citeproc` and not `get_pubmed_citeproc` to use the [Literature Citation Exporter](https://api.ncbi.nlm.nih.gov/lit/ctxp). We can migrate `get_pubmed_citeproc` in the future, but it's not pressing.

Once merged, we should update rootstock and release Manubot 0.2.1 (I will start preparing release notes).

dhimmel:
Ah interesting. When querying `https://api.ncbi.nlm.nih.gov/lit/ctxp/v1/pmc/?format=csl&id=` the following JSON is returned:

```json
{
    "status": "error",
    "reason": {
        "id": [
            [
                "This field is required."
            ]
        ]
    }
}
```

However, when querying an invalid id such as `https://api.ncbi.nlm.nih.gov/lit/ctxp/v1/pmc/?format=csl&id=3`, the following XML is returned:

```xml
<pub-one-record-set xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:p1="http://pubmed.gov/pub-one" p1:error="PubOne failure: empty XML."/>
```

dhimmel:
@agitter I made the error checking and messaging more robust. Take a look again?

PR #86, opened 2018-12-07
Wikidata zotero

dhimmel:
Requires rebase on other PRs post merge

Closes https://github.com/greenelab/manubot/issues/67
Supersedes https://github.com/greenelab/manubot/pull/78


tests/test_zotero.py
```diff
@@ -82,3 +83,17 @@ def test_search_query():
     identifier = 'isbn:9781339919881'
     zotero_data = search_query(identifier)
     assert zotero_data[0]['title'].startswith('The hetnet awakens')
+
+
+def test_get_wikidata_citeproc():
+    """
+    Test metadata extraction from https://www.wikidata.org/wiki/Q50051684
+
+    Author ordering is not properly set by the Wikidata translator
```
agitter: This seems like a big problem.  Users won't be aware that the author ordering could be incorrect.  Do we want to wait to merge until the underlying issue can be resolved?

dhimmel: Currently there's an open PR to fix this: https://github.com/zotero/translators/pull/1794. I'm in favor of merging now with the issue, since I want to reorganize the tests, and it'd be easiest with no open PRs.

By time this makes it into a release or manubot-rootstock, I expect we'll have fixed the upstream issue.

agitter: Thanks for clarifying.  I'm more comfortable merging if there is an upstream PR in progress.

Should we create a new test for the author order and mark it with xfail until the upstream problem is fixed?  That could help ensure we don't lose track of this behavior if https://github.com/zotero/translators/pull/1794 is not merged quickly.

dhimmel: done in https://github.com/greenelab/manubot/pull/86/commits/0512371d2b0eec3b1359a383cbc662b11291f9a1

PR #90, opened 2018-12-19
translation-server web query: force single result

dhimmel:
Refs https://github.com/zotero/translation-server/issues/65

### DIFF
--- a/manubot/cite/tests/test_isbn.py
+++ b/manubot/cite/tests/test_isbn.py
@@ -33,13 +33,12 @@ def test_get_isbn_citeproc_citoid_not_found():
         get_isbn_citeproc_citoid('9781439566039')
 
 
-@pytest.mark.xfail(reason='https://github.com/zotero/translation-server/issues/67')
 def test_get_isbn_citeproc_zotero_with_note_issue():
     """
     translation-server returns two metadata records for this ISBN.
-    The second has itemType=note and causes CSL export to fail.
+    The second has itemType=note and previousely caused CSL export to fail.
     https://github.com/zotero/translation-server/issues/67
     """
     isbn = '9780262517638'
     csl_item = get_isbn_citeproc_zotero(isbn)
-    assert csl_item['author'][0]['given'] == 'Suber'
+    assert csl_item['author'][0]['family'] == 'Suber'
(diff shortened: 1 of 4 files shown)


agitter:
Because there are so many types of Zotero translators, do you think we need more test cases to help with regression testing on various frequently-used translators?

dhimmel:
> Because there are so many types of Zotero translators, do you think we need more test cases to help with regression testing on various frequently-used translators?

Possibly although I believe translators often include tests in their definitions in https://github.com/zotero/translators. Also I think translation-server has some [tests](https://github.com/zotero/translation-server), although does not appear to use CI. @dstillman, is that correct?

Anyways, I think we should add more translation-server tests here for safety, but shouldn't get carried away, as this is not really the right location. I'd rather help the upstream repositories have the right tests than duplicating them here.

dstillman:
> Also I think translation-server has some tests, although does not appear to use CI. @dstillman, is that correct?

I've just set up Travis for translation-server.

Translator tests are at https://translator-tests.zotero.org/. (These tests were recently reinstated using the new translation server, so there are currently a larger number of failures than there have been at other times, but in many cases it will still fall back to other translators.)

agitter:
> Anyways, I think we should add more translation-server tests here for safety, but shouldn't get carried away, as this is not really the right location. I'd rather help the upstream repositories have the right tests than duplicating them here.

> We should however add regression tests whenever we notice an issue.

Yes, I agree with this.  For instance, GitHub repository URLs and Manubot manuscript URLs are frequently-cited, so those could be good test cases.

dhimmel:
> Yes, I agree with this. For instance, GitHub repository URLs and Manubot manuscript URLs are frequently-cited, so those could be good test cases.

Added in https://github.com/greenelab/manubot/pull/90/commits/c3b3833db2c1f612cdb14d955c7595052f31a3c7.

In https://github.com/greenelab/manubot/pull/90/commits/da32e7c4849fb920b02da88a390f7abf4857045c, I added /search tests for PMIDs, arXiv IDs, and DOIs. Currently, we don't use these, but I thought it'd be good to test to make sure our translation-server instance is working properly.

PR #93, opened 2019-01-23
Support shortDOI citation

### LINKED ISSUE #92: shortDOI citation support
Supporting citations of shortDOIs could be a nice feature. I see the major benefits as saving characters for excessively long DOIs and enabling a way to directly cite DOIs with forbidden characters such as `10.1016/S0933-3657(96)00367-3`. At least one user has attempted this workaround in https://github.com/Benjamin-Lee/deep-rules/pull/117/commits/c76ee4d20fa3f45935194b43b495b8159cbd30b4 / https://github.com/Benjamin-Lee/deep-rules/pull/117#issuecomment-456227977.

It seems like we could support a few different syntaxes:

1. `@doi:10/b6vnmd`
2. `@doi:b6vnmd`
3. `@shortdoi:b6vnmd`
4. `@shortdoi:10/b6vnmd`

I didn't see much shortDOI documentation online, but there is some provided when viewing a [shortened DOI result](http://web.archive.org/web/20190123160439/http://shortdoi.org/10.1016/S0933-3657(96)00367-3)

```
Your request was processed. The previously-created shortcut for 10.1016/S0933-3657(96)00367-3 is the handle:
10/b6vnmd
The shortcut HTTP URI is:
http://doi.org/b6vnmd
This shortcut will return the same results as http://dx.doi.org/10.1016/S0933-3657(96)00367-3, and doi:10/b6vnmd can be used in place of doi:10.1016/S0933-3657(96)00367-3.
```

Given the documentation, it seems that option 1 is the most canonical method. However, method 3 & 4 could help avoid user confusion.

***

Note that DOI content negotiation for crossref DOIs appears to work:

```
curl --location --header "Accept: application/vnd.citationstyles.csl+json" https://doi.org/b6vnmd
```

dhimmel:
Closes #92

### DIFF
--- a/manubot/cite/util.py
+++ b/manubot/cite/util.py
@@ -36,10 +36,13 @@ citation_pattern = re.compile(
 
 def standardize_citation(citation):
     """
-    Standardize citation idenfiers based on their source
+    Standardize citation identifiers based on their source
     """
     source, identifier = citation.split(':', 1)
     if source == 'doi':
+        if identifier.startswith('10/'):
+            from manubot.cite.doi import expand_short_doi
+            identifier = expand_short_doi(identifier)
         identifier = identifier.lower()
     if source == 'isbn':
         from isbnlib import to_isbn13
@@ -51,6 +54,7 @@ regexes = {
     'pmid': re.compile(r'[1-9][0-9]{0,7}'),
     'pmcid': re.compile(r'PMC[0-9]+'),
     'doi': re.compile(r'10\.[0-9]{4,9}/\S+'),
+    'shortdoi': re.compile(r'10/[a-zA-Z0-9]+'),
     'wikidata': re.compile(r'Q[0-9]+'),
 }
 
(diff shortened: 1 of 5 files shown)


agitter:
Do you envision expanding the short DOIs before the rest of the citation processing?  I'm considering the behavior if a user provides the shortDOI in one citation and the corresponding DOI in another.  Can Manubot treat these as the same reference?

My preference for the `shortdoi:` versus `doi:` prefix is whichever makes it easier to match citations to the same reference in the scenario above.

dhimmel:
> Do you envision expanding the short DOIs before the rest of the citation processing?

Yes that is the design I think is best. The shortDOI would be expanded by time it's converted to a `standard_citation`. This would make it so `@doi:10/b6vnmd` and `@doi:10.1016/S0933-3657(96)00367-3` cite the same reference.

dhimmel:
> if it isn't time sensitive it will take me a little while

not a problem. take your time.

manubot/cite/util.py
```diff
@@ -36,21 +36,38 @@
...
     if source == 'doi':
+        if identifier.startswith('10/'):
+            from manubot.cite.doi import expand_short_doi
+            try:
+                identifier = expand_short_doi(identifier)
+            except Exception as error:
+                # If DOI shortening fails, return the unshortened DOI.
+                # DOI metadata lookup will eventually fail somewhere with
+                # appropriate error handling, as opposed to here.
+                logging.error(
```
agitter: This error message prints twice if I try using a shortDOI that is invalid but passes the regex.  Is that expected?

`manubot cite doi:10/98p1`

dhimmel: That is because `standardize_citation` was being called by `citation_to_citeproc` to assert that the input citation is already standardized:

https://github.com/manubot/manubot/blob/1b3ac66de9220e0b7496128a4a45e444b6d4edb1/manubot/cite/util.py#L216-L220

I am not sure whether we really need to do that check, but the easiest solution was to cache `standardize_citation` in https://github.com/manubot/manubot/pull/93/commits/8ed947318bf483769de78b5ec490337e87d23221. This should have potential performance improvements as well.

PR #96, opened 2019-02-01
Relocate repo from greenelab/manubot to manubot/manubot

dhimmel:
Closes https://github.com/manubot/manubot/issues/94


dhimmel:
@agitter ready for review.

I don't think the PyPI deployment should break, but we'll have to wait and see until our next release.

Issue #13, opened 2017-09-07
Do not extract or replace citation strings inside code or code_blocks


dhimmel:
Right now our manuscript parsing implementation is unaware of code and code blocks. As a result, citation strings inside of of code or code blocks will be extracted and replaced. This is not the desired behavior -- text formatted as code should not get modified.

For example:

```markdown
@tag:citation should get extracted and replaced.
However, `@tag:citation` should not.
```

To solve this issue, we need to be able to easily separate parts of the source markdown that are code or code blocks.

dhimmel:
I see the general solution to this issue as switching our citation processing to use a pandoc filter. This is underway in https://github.com/manubot/manubot/pull/99.

PR #99, opened 2019-03-08
Create a pandoc-manubot-cite filter for pandoc

dhimmel:
Create a pandoc filter providing manubot's cite-by-ID functionality.

Todo:

- [x] logging exception to nonzero exit status (off by default)
- [x] caching (off by default)
- [x] export `references.json` to file in addition to pandoc metadata?


dhimmel:
As of https://github.com/manubot/manubot/pull/99/commits/87d1e3c04bb152c0457cb72b45eba01193f1978d, I've implemented a proof of concept pandoc filter that extracts `citationId`s from pandoc's AST (abstract syntax tree) and replaces them with manubot `citation_id`s. The primary remaining step would be to generate CSL and store it in the appropriate pandoc metadata fields.

Since it seems feasible to create this filter, I thought I'd open the idea to discussion of its merits. The main benefits are:

1. the filter could massively increase the userbase of manubot's cite-by-id functionality.
2. the filter would use pandoc's parsing of documents to extract citations, helping fix bugs like how currently citations in code elements are modified (see https://github.com/manubot/manubot/issues/13).

The downsides are:

1. Increased support burden. More research is needed how the plugin will interact with common pandoc commands, especially those using bibtex citations.
2. Currently, we use a less stringent method for matching citations than pandoc:

https://github.com/manubot/manubot/blob/81394c0a8153fa3ad41ee3cb987227c8ea043125/manubot/cite/util.py#L18-L35

    If we want to switch from using `manubot process` to using a pandoc filter for citation-by-id for manubot manuscripts, then we will break some existing citation strings. See also discussion at https://github.com/manubot/rootstock/issues/2#issuecomment-317214407, where @tarleb initially suggested using a pandoc filter for this purpose. One aspect will be how often are persistent identifiers invalid pandoc citations due to forbidden characters.

manubot/pandoc_filter/cite.py
```diff
@@ -0,0 +1,102 @@
...
+}
+
+
+def parse_args():
+    """
+    Read command line arguments
+    """
+    parser = argparse.ArgumentParser(description='Pandoc filter for citation by persistent identifier')
+    parser.add_argument('target_format')
+    parser.add_argument('--pandocversion', help='The pandoc version.')
```
dhimmel: @tomduck, I copied this `--pandocversion` argument [from](https://github.com/tomduck/pandoc-fignos/blob/44776d885b8101d9d6aeb6845b17ad1625e88e6f/pandoc_fignos.py#L63) `pandoc-fignos`. Do you know what it's for and whether it's needed? It doesn't seem to me that pandoc 2.5 supplies this argument to filters.

manubot/pandoc_filter/cite.py
```diff
@@ -0,0 +1,166 @@
+"""
+Preliminary testing command:
+
+pandoc \
+  --to=plain \
+  --standalone \
+  --filter=pandoc-manubot-cite \
+  --filter pandoc-citeproc \
+  manubot/pandoc_filter/tests/input-with-cites.md
+
```
slochower: Is there a difference between `--filter=` and `--filter ` (with a space)?

dhimmel: No

dhimmel:
Note to consider the syntax used by `pandoc-url2cite` for defining citekey aliases using the explicit link reference definitions ([source code](https://github.com/phiresky/pandoc-url2cite/blob/b28374a9a037a5ce1747b8567160d8dffd64177e/index.ts#L118-L236)).

```
This is a sentence [@citekey1; @citekey2].

[@citekey1]: https://github.com/manubot/manubot/pull/99
[@citekey2]: doi:10/...
```

manubot/pandoc/cite.py
```diff
@@ -0,0 +1,177 @@
...
+            # save info to citekeys and remove from paragraph
+            citekey = elem.content[0].citations[0].id  # differs from pandoc-url2cite
+            citekey_aliases = global_variables["citekey_aliases"]
+            if citekey in citekey_aliases:
+                logging.warning(f"duplicate citekey {citekey}")
+            citekey_aliases[citekey] = destination.text
+            # found citation, add it to citekeys and remove it from document
+            elem.content = elem.content[space_index + 1 :]
+        if len(elem.content) > 0 and type(elem.content[0]) == pf.SoftBreak:
+            elem.content.pop(0)
```
dhimmel: @phiresky following our discussing in https://github.com/phiresky/pandoc-url2cite/issues/1, I was curious to see if we could add the link-reference-syntax for defining citekey aliases to our in-progess pandoc filter.

I based this function heavily off of your [`extractCiteKeys`](https://github.com/phiresky/pandoc-url2cite/blob/b28374a9a037a5ce1747b8567160d8dffd64177e/index.ts#L118-L152) function. I noticed your code is [AGPL licensed](https://github.com/phiresky/pandoc-url2cite/blob/master/LICENSE.md). For this repo, we use the BSD 3-Clause License, which is more permissive and hence incompatible with AGPL. Are you okay with us reusing your logic here under this repository's BSD license? We would retain the link in the documentation to your package and elsewhere as appropriate.

phiresky: I hereby release the code in the extractCiteKeys function to the public domain. Feel free to use it in BSD 3-Clause licensed code. I guess I wouldn't have any legal claim since you rewrote it anyways though.

My implementation might have some robustness issues, so don't trust it. It would be best if it was exactly the same as the normal markdown syntax but my implementation was enough for my purposes for now. (Defining link references in random places within paragraphs is ugly anyways). 

This would be another issue that we might want to open in the pandoc repo: the suggestion to somewhat re-unify the syntax for citations and links regarding the handling of both `[@test](http://...)` and `[@test] .... [@test]: http://`. In my opinion the parsing should be as close as possible to normal markdown when citation parsing is active (e.g. to commonmark or to `pandoc --format markdown-citations`). Both for familiarity for new pandoc users as well as for writing "polyglot" pandoc markdown code that "mostly" works with other processors (e.g. github markdown).

dhimmel: > Feel free to use it in BSD 3-Clause licensed code. I guess I wouldn't have any legal claim since you rewrote it anyways though.

Thanks! I would say I more translated it than completely rewrote it... so probably a gray area in terms of whether copyright applies.

> In my opinion the parsing should be as close as possible to normal markdown when citation parsing is active

So you think `pandoc --format markdown+citations` should treat `[@test](http://...)` like a hyperlink on `@test` without processing `@test` as a citation? Or process `@test` and also hyperlink it?

phiresky: I think it should still treat it as a citation and either ignore what's written in brackets or somehow try to interpret it as cite data or just add it as metadata to the citation in the AST... or process it but also link it like you said. Not sure. Mainly I think it should not write `(http://...)` as literal text to the output like it currently does since that also doesn't happen in non-pandoc markdown.

Same for `[@test][bla]   \n [bla]: 123` which is `<a href=123>@test</a>` in markdown but `<span class="citation" data-cites="test">[@test]</span><a href="123">bla</a>`.

dhimmel:
This is ready for review. CC @agitter 

I am thinking we can merge this this feature and continue to prototype the `pandoc-manubot-cite` filter before switching rootstock to use it.

dhimmel:
@agitter would be great for you to take a look at this. Would like to move forward with merging.

agitter:
Okay, I'll try to make time Friday or Saturday.

manubot/pandoc/cite_filter.py
```diff
@@ -0,0 +1,219 @@
...
+
+    Uses markdown's link reference syntax to define citekey aliases (tags)
+    https://spec.commonmark.org/0.29/#link-reference-definitions
+    """
+    if type(elem) != pf.Para:
+        return
+    while (
+        len(elem.content) >= 3
+        and type(elem.content[0]) == pf.Cite
+        and len(elem.content[0].citations) == 1  # differs from pandoc-url2cite
```
agitter: How does it differ?  Can you comment on the difference here and below?

manubot/pandoc/cite_filter.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+    Based on TypeScript implementation by phiresky at
+    https://github.com/phiresky/pandoc-url2cite/blob/b28374a9a037a5ce1747b8567160d8dffd64177e/index.ts#L118-L152
+
+    Uses markdown's link reference syntax to define citekey aliases (tags)
+    https://spec.commonmark.org/0.29/#link-reference-definitions
+    """
+    if type(elem) != pf.Para:
+        return
+    while (
```
agitter: It's hard to follow this loop without knowing the structure of `pf.Para`.  Is this applied only on Markdown link references?  Or it's extracting citekeys and their aliases from the entire document?

manubot/pandoc/cite_filter.py
```diff
@@ -0,0 +1,219 @@
...
+            f"Expected metadata.citekey-aliases to be a dict, "
+            f"but received a {citekey_aliases.__class__.__name__}. Disregarding."
+        )
+        citekey_aliases = dict()
+
+    global_variables["citekey_aliases"] = citekey_aliases
+
+    manuscript_citekeys = set(global_variables["manuscript_citekeys"])
+
+    doc.walk(_get_reference_link_citekey_aliases)
```
agitter: What does `doc.walk` do?  Does it traverse the AST and apply the provided action at all elements (tree nodes)?

manubot/pandoc/cite_filter.py
```diff
@@ -0,0 +1,219 @@
...
+    if not isinstance(citekey_aliases, dict):
+        logging.warning(
+            f"Expected metadata.citekey-aliases to be a dict, "
+            f"but received a {citekey_aliases.__class__.__name__}. Disregarding."
+        )
+        citekey_aliases = dict()
+
+    global_variables["citekey_aliases"] = citekey_aliases
+
+    manuscript_citekeys = set(global_variables["manuscript_citekeys"])
```
agitter: Why does this need to be called before the two `walk` calls?

manubot/pandoc/cite_filter.py
```diff
@@ -0,0 +1,219 @@
...
+        pathlib.Path(requests_cache_path).parent.mkdir(parents=True, exist_ok=True)
+    csl_items = generate_csl_items(
+        citekeys=standard_citekeys,
+        manual_refs=manual_refs,
+        requests_cache_path=doc.get_metadata("manubot.requests-cache-path"),
+        clear_requests_cache=doc.get_metadata("manubot.clear-requests-cache", False),
+    )
+    output_bibliography = doc.get_metadata("manubot.output-bibliography")
+    write_csl_json(csl_items, output_bibliography)
+    doc.metadata["bibliography"] = []
```
agitter: What is this line doing?

manubot/pandoc/tests/test_cite_filter/input.md
```diff
@@ -0,0 +1,45 @@
+---
+# yaml_metadata_block with pandoc metadata
+citekey-aliases:
+  tag:meta-review: url:https://greenelab.github.io/meta-review/
```
agitter: Is this `tag:meta-review:` syntax temporary to be backwards compatible?  Or is this how you envision using tags/aliases in the future?

manubot/pandoc/tests/test_cite_filter/output.txt
```diff
@@ -0,0 +1,57 @@
+Testing markdown document for pandoc-manubot-cite filter
+
+This is a sentence with one citation [1].
+
+This is a sentence with many citations [1,2].
+
+    Citations in code blocks should not be modified [@pmid:20170387].
```
agitter: This is very nice to have!

manubot/pandoc/tests/test_cite_filter/input.md
```diff
@@ -0,0 +1,45 @@
...
+
+Citing pre-defined citekey aliases [@tag:meta-review; @tag:deep-review].
+
+Citing @raw:dongbo-conversation whose metadata is already in pandoc's reference metadata.
+
+Citing works whose reference metadata is provided via `--bibliography` [@arxiv:1407.3561; @tag:paper-of-future].
+
+## References
+
+::: {#refs}
```
agitter: This is the Pandoc syntax for outputting the reference list?

manubot/pandoc/tests/test_cite_filter/input.md
```diff
@@ -0,0 +1,45 @@
...
+Defining citekeys with the link reference syntax [@tag:issue; @tag:bad-doi; @tag:bad-url].
+
+[@tag:issue]: url:https://github.com/manubot/manubot/pull/189
+[@tag:bad-doi]: doi:10.1016/S0022-2836(05)80360-2
+
+[@tag:bad-url]: url:https://openreview.net/forum?id=HkwoSDPgg
+
+Citing pre-defined citekey aliases [@tag:meta-review; @tag:deep-review].
+
+Citing @raw:dongbo-conversation whose metadata is already in pandoc's reference metadata.
```
agitter: Do we see this being used?  Is the idea that a valid Pandoc filter can't ignore valid reference metadata if someone were to use it outside the typical Manbot workflow?

manubot/process/tests/test_process_command.py
```diff
@@ -27,14 +30,18 @@ def test_example_manuscript(manuscript):
         "--output-directory",
         str(manuscript_dir.joinpath("output")),
     ]
+    if skip_citations:
+        args.append("--skip-citations")
```
agitter: Do you want to test that skip citations actually skips the citation processing?

manubot/pandoc/tests/test_cite_filter/input.md
dhimmel: The idea is that we should support the standard pandoc methods for defining reference metadata in addition to any manubot magic. That means we need to support `metadata.bibliography` (paths to bibliography files) and `metadata.references` (CSL JSON style reference metadata), since these are the standard pandoc ways to specify reference metadata.

manubot/pandoc/tests/test_cite_filter/input.md
dhimmel: yes as per https://pandoc.org/MANUAL.html

manubot/pandoc/tests/test_cite_filter/input.md
dhimmel: Backwards compatible syntax that is required until we implement #192. After we implement #192 then you will no longer need to prefix with `tag:`. But for now it's needed.

manubot/command.py
```diff
@@ -72,10 +72,15 @@ def add_subparser_process(subparsers):
         "like `--template-variables-path=namespace=path_or_url`. "
         "Namespaces must match the regex `[a-zA-Z_][a-zA-Z0-9_]*`.",
     )
+    parser.add_argument(
+        "--skip-citations",
+        action="store_true",
+        help="Skip citation and reference processing.",
```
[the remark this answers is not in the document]
dhimmel: Not sure, but `manubot process` never does any fig/table reference processing since those are left untouched for the pandoc-xnos filters.

manubot/pandoc/cite_filter.py
dhimmel: comment was relic. the difference is because panflute syntax is different, but the functionality is the same (as far as I can tell)

manubot/pandoc/cite_filter.py
dhimmel: good call. this line is completely inert as far as i can tell

manubot/pandoc/cite_filter.py
dhimmel: > Walk through the element and all its children (sub-elements), applying the provided function action.

manubot/pandoc/cite_filter.py
dhimmel: If the filter receives `metadata.bibliography` files, we load the references from those files. Therefore, we don't want pandoc-citeproc to re-read the references in those files. So we remove them from the metadata.

manubot/process/tests/test_process_command.py
dhimmel: Not really because I view `manubot process` as a command we're phasing out and test takes time to write. Relying on https://github.com/manubot/rootstock/pull/300 to test `--skip-citations` actually works.

manubot/pandoc/cite_filter.py
dhimmel: comments added.

> Is this applied only on Markdown link references? Or it's extracting citekeys and their aliases from the entire document?

The former

README.md
```diff
@@ -190,6 +191,39 @@ optional arguments:
...
+positional arguments:
+  target_format
+
+optional arguments:
+  -h, --help         show this help message and exit
+  --input [INPUT]    path read JSON input (defaults to stdin)
+  --output [OUTPUT]  path to write JSON output (defaults to stdout)
+```
+
+Other pandoc filters exist that do something similar:
```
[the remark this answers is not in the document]
dhimmel: personally I don't care much about the capitalization in our docs / help messages / comments. Consistency is nice, but reducing the cognitive overhead of writing docs. Also sometimes pandoc refers to the `pandoc` command which is lowercase

manubot/process/tests/test_process_command.py
dhimmel: `--skip-citations` is working in https://github.com/manubot/rootstock/runs/387673500 / https://github.com/manubot/rootstock/commit/49dcfcdef8c9bf97cc4f33b561657ecec1aea319 which relies on `citation-tags.tsv`.

dhimmel:
https://github.com/manubot/manubot/pull/99/commits/be5bc4e8d9abee7032ac70d54d720d52e802944c is working. see https://travis-ci.com/manubot/rootstock/builds/144231641#L855

dhimmel:
@agitter I'm going to merge and address additional feedback in subsequent PRs.

agitter:
> I'm going to merge and address additional feedback in subsequent PRs.

Sounds good to me.  I'll look over your responses and new commits here later in the week.

PR #110, opened 2019-05-18
Testing manuscript to evaluate combinations of missing CSL JSON fields

dhimmel:
Generate CSL Items with all combinations of fields
To test CSL bibliography rendering with different fields missing.

refs https://github.com/manubot/rootstock/issues/155


slochower:
This is looking great!

PR #111, opened 2019-05-22
setup.py: use package_data not MANIFEST to include data files

dhimmel:
In https://github.com/manubot/manubot/pull/51, we could not get `package_data` in `setup.py` to work on Windows. We reported the issue in https://github.com/pypa/setuptools/issues/1459 and switched to using `MANIFEST.in`. The downside of MANIFEST is that it creates an extra file in the root directory of the repository for a rather obscure technical purpose.

In https://github.com/pypa/setuptools/issues/1459#issuecomment-493506713, @aschmied hypothesized `package_data` did not work for us because we passed a string rather than a list. This PR switches back to the `package_data` with the correct syntax.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,0 @@
-include manubot/cite/*.lua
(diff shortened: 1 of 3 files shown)


dhimmel:
Travis builds are currently not running due to a [GitHub outage](https://www.githubstatus.com/incidents/nf4gcjtcsqdb), but the passing AppVeyor Windows build is a good early indication this solution works.

PR #113, opened 2019-05-23
Internal citation nomenclature & design refactor

dhimmel:
This behavior seems most consistent with Pandoc and CSL behavior. `@` is used to signify a citation, but should not be part of the actual citation identifier.


agitter:
I can review when this is ready

dhimmel:
This PR updates the columns of `citation_df` to make it more clear what has been done to the citation id at each step. The new column names are:

- manuscript_id: citation ids extracted from the manuscript content files.
- detagged_id: manuscript_id but with tag citations dereferenced
- standard_id: detagged_id standardized
- short_id: standard_id hashed to create a short base

I would also like to agree on a more explicit sequence of processing steps for citation_ids from the manuscript text, citation_ids from the manual references, and citation_ids from `manubot cite`.

Do we want to `infer_citation_prefix` on citation_ids from the manuscript text for example? This would accommodate users who want to use their existing IDs as raw citaitons without having to prefix `raw:`. It also would convert `DOI:` to `doi:`.

agitter:
> Do we want to infer_citation_prefix on citation_ids from the manuscript text for example?

The major advantage would be more flexible support for `raw` ids.  When paired with the recent changes in #104, would that be sufficient to support citations to external references in Manubot manuscripts?  That seems like a good enhancement to me, as does the case insensitivity for the persistent id types.

Are there any disadvantages we can think of?

slochower:
> The major advantage would be more flexible support for raw ids. When paired with the recent changes in #104, would that be sufficient to support citations to external references in Manubot manuscripts? That seems like a good enhancement to me, as does the case insensitivity for the persistent id types.

Is the idea that you could specify a `bib` file with `bibkey`s (say, if you have these lying around from previous work or whatever) and then use `raw:bibkey` with Manubot to cite references in the `bib` file?

dhimmel:
> Is the idea that you could specify a bib file with bibkeys (say, if you have these lying around from previous work or whatever) and then use raw:bibkey with Manubot to cite references in the bib file?

You can already do this as of https://github.com/manubot/manubot/commit/37a9c69ad8d74d51ee3cff7779b2b6d303629c07. The question is whether we should support citing `@bibkey` in the manuscript in addition to the current `@raw:bibkey`.

slochower:
> You can already do this as of 37a9c69. The question is whether we should support citing @bibkey in the manuscript in addition to the current @raw:bibkey.

If it's just a stylistic thing, personally `@bib:bibkey` would probably make more sense? I think the idea of `@raw` probably isn't obvious.

dhimmel:
> If it's just a stylistic thing, personally `@bib:bibkey` would probably make more sense? I think the idea of `@raw` probably isn't obvious.

`raw` here refers to any citation whose metadata is not from one of the supported sources. Therefore, we use it for all IDs from manual references files (that could be in JSON CSL format, bibtex, or many other formats). Creating the special case where citations derived from a `.bib` file use `bib:` not `raw:` would add a lot of complexity and create a special case that doesn't need to exist.

slochower:
>raw here refers to any citation whose metadata is not from one of the supported sources. Therefore, we use it for all IDs from manual references files (that could be in JSON CSL format, bibtex, or many other formats). Creating the special case where citations derived from a .bib file use bib: not raw: would add a lot of complexity and create a special case that doesn't need to exist.

Fair enough. 

The context I had in mind is actually the inverse "problem" and distinct from this. Composing with Manubot and using a mix of known sources (e.g., `@doi:whatever`) and `raw` ones. Then exporting to `.tex` plus `.bib`. Current `.tex` export via `pandoc` won't result in a separate `bib` file. We could convert `references.json` to `.bib` for LaTeX export, but again, this is a totally different Issue.

agitter:
> One issue with this is that it could make it harder to detect mistake IDs like doid:10/blah, since they would get converted to raw:doid:10/blah

I had been thinking about the lines in `is_valid_citation_string` that split on `:` and use that to set the source.  I see now how calling `infer_citation_prefix` first could result in `raw:doid:10/blah`.  That does seem undesirable.

Do we want to make the behavior more restrictive by only prepending `raw:` if the ID does not contain `:` already?  That could make valid keys in `.bib` files, etc. fail, however.

dhimmel:
> Do we want to make the behavior more restrictive by only prepending `raw:` if the ID does not contain `:` already?

This is a good idea. There are likely going to be some IDs that have `:`, like this automatically generated ID by zbib / zotero:

https://github.com/manubot/manubot/blob/37a9c69ad8d74d51ee3cff7779b2b6d303629c07/manubot/pandoc/tests/bibliographies/bibliography.bib#L15

What about if we still preprend `raw:` but log a warning if the citation contains `:`? We could also consider a command line option like `--infer-raw-citations` to make this an opt-in feature.

dhimmel:
On second thought, I think it will be too much for this PR to add `raw` prefixing of manuscript-derived citation IDs. But we should consider it in the future.

agitter:
>  I think it will be too much for this PR to add raw prefixing of manuscript-derived citation IDs.

Okay.  This PR is getting fairly large already.  Adding a command line option in the future may be the right choice to toggle between a strict and permissive mode.

dhimmel:
# Meta Review Integration Test

I updated `manubot` to https://github.com/manubot/manubot/pull/113/commits/f19a8a00967ea5d440635b40c974c7ac65a905f8 and rebuilt the meta-review locally, which produced the following PDF [manuscript.pdf](https://github.com/manubot/manubot/files/3238493/manuscript.pdf).

`output/references.json` contained notes like the following:

```
    "note": "This CSL JSON Item was automatically generated by Manubot v0.2.3 using citation-by-identifier.\nstandard_id: doi:10.1038/d41586-017-08493-x",

    "note": "This CSL JSON Item was loaded by Manubot v0.2.3 from a manual reference file.\nmanual_reference_filename: manual-references.json\nstandard_id: raw:dongbo-conversation",
```

I copied `output/references.json` to `content/manual-references-2019-05-30.json` and confirmed the build was now much faster.

The new `output/references.json` had notes like:

```
    "note": "This CSL JSON Item was automatically generated by Manubot v0.2.3 using citation-by-identifier.\nstandard_id: doi:10.1038/d41586-017-08493-x\nThis CSL JSON Item was loaded by Manubot v0.2.3 from a manual reference file.\nmanual_reference_filename: manual-references-2019-05-30.json\noriginal_id: 9IrsqXRa",

    "note": "This CSL JSON Item was loaded by Manubot v0.2.3 from a manual reference file.\nmanual_reference_filename: manual-references.json\nstandard_id: raw:dongbo-conversation",
```

Note that `content/manual-references-2019-05-30.json` was loaded before `manual-references.json`, such that the CSL JSON in `manual-references.json` overrided the former. Not sure whether we want to tweak the sorting in:

https://github.com/manubot/manubot/blob/351f4bac474d478b59283aa57afb8f7b625c0e80/manubot/process/manuscript.py#L32

@agitter ready for review.

manubot/cite/util.py
```diff
@@ -267,12 +276,12 @@ def citation_to_citeproc(citation, prune=True):
 
     note_text = f'This CSL JSON Item was automatically generated by Manubot v{manubot_version} using citation-by-identifier.'
     note_dict = {
-        'standard_citation': citation,
```
agitter: This will affect reading and writing CSL JSON files, right?  Is there any risk of introducing incompatibilities with existing CSL JSON?  It looks like the answer is no because `csl_item_set_standard_id` looks for `standard_id` and then `standard_citation`.

manubot/process/manuscript.py
```diff
@@ -37,18 +37,18 @@ def get_text(directory):
     return '\n\n'.join(name_to_text.values()) + '\n'
 
 
-def replace_citations_strings_with_ids(text, string_to_id):
+def update_manuscript_citations(text, old_to_new):
```
agitter: `old_to_new` is a generic name.  Would `old_to_short` be accurate and more descriptive?

dhimmel: I did old_to_new because this function could be used to change citations to a new format other than short. In other words, function is more general, but our use case is specific.

dhimmel:
> Before merging, do we need a new issue to track ongoing discussion about the sequence of processing steps

Created https://github.com/manubot/manubot/issues/114.

> I don't have any strong preference regarding the sort order of manual references files. As long as it it stale and deterministic, it should be workable.

GitHub sorts by filename without special treatment of the extension ([example](https://github.com/greenelab/meta-review/tree/84d00c76047928d4ca424e60e3ce49cf132d3918/content)):

> image

This matches our current behavior and the behavior of `ls`, so let's stick with it.

manubot/cite/util.py
dhimmel: Adding the note was recently added in https://github.com/manubot/manubot/pull/104, which has not been included in rootstock, so I don't think there will be any CSL JSON notes that use `standard_citation`. Also I don't think there are many cases where it would matter.

PR #116, opened 2019-06-04
Python 3.7 on Travis & fix YAMLLoadWarning

dhimmel:


dhimmel:
The Python 3.7 test will let us see warnings like [these ones](https://travis-ci.com/manubot/manubot/jobs/205309499#L541), and will be helpful for knowing when we can upgrade rootstock to Python 3.7. Refs https://github.com/manubot/rootstock/pull/227#issuecomment-498486187

PR #118, opened 2019-06-25
Extend documentation of installation

nichtich:
Added installation via PyPI. Installation should be explained before usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,23 @@ To cite the Manubot project or for more information on its design and history, s
 Daniel S. Himmelstein, Vincent Rubinetti, David R. Slochower, Dongbo Hu, Venkat S. Malladi, Casey S. Greene, Anthony Gitter<br>
 _Manubot Preprint_ (2019) <https://greenelab.github.io/meta-review/>
 
+## Installation
+
+Install the latest release version:
+
+```sh
+pip install manubot
+```
+
+or install a version specified by a git commit hash:
+
+```sh
+COMMIT=33e512d21218263423de5f0d127aac4f8635468f
+pip install git+https://github.com/manubot/manubot@$COMMIT
+```
+
+Use the `--upgrade` argument to reinstall `manubot` with a different commit hash.
+
 ## Usage
 
 Installing the python package creates the `manubot` command line program.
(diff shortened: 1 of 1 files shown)


agitter:
@nichtich thanks for the contribution

@dhimmel in the installation instructions do we also want to point out that `manubot process` as used in the rootstock build script requires coordinating the Manubot package version with versions of other packages and files?  That is, a note similar to https://github.com/manubot/rootstock/pull/232

slochower:
> @dhimmel in the installation instructions do we also want to point out that manubot process as used in the rootstock build script requires coordinating the Manubot package version with versions of other packages and files? That is, a note similar to manubot/rootstock#232

I think that would be good to address, given that it has already popped up.

README.md
```diff
@@ -15,6 +15,23 @@ To cite the Manubot project or for more information on its design and history, s
 Daniel S. Himmelstein, Vincent Rubinetti, David R. Slochower, Dongbo Hu, Venkat S. Malladi, Casey S. Greene, Anthony Gitter<br>
 _Manubot Preprint_ (2019) <https://greenelab.github.io/meta-review/>
 
+## Installation
+
+Install the latest release version:
```
dhimmel: Can we change this sentence to:

```
Install the latest release version [from PyPI](https://pypi.org/project/manubot/):
```

README.md
```diff
@@ -15,6 +15,23 @@ To cite the Manubot project or for more information on its design and history, s
...
+```
+
+or install a version specified by a git commit hash:
+
+```sh
+COMMIT=33e512d21218263423de5f0d127aac4f8635468f
+pip install git+https://github.com/manubot/manubot@$COMMIT
+```
+
+Use the `--upgrade` argument to reinstall `manubot` with a different commit hash.
```
dhimmel: I think it's possible that we could just add `--upgrade` to both `pip install` commands.

From https://github.com/conda/conda/pull/8542:

> Adding the flag here causes it to always be used, even for `conda env create`, but in that case the -U has no effect because the package is not already installed.

This way the commands work as intended regardless of whether `manubot` is already installed. We could add a comment or note that `--upgrade` ensures existing installations are replaced with the specified version as neccessary.

README.md
```diff
@@ -15,6 +15,23 @@ To cite the Manubot project or for more information on its design and history, s
 Daniel S. Himmelstein, Vincent Rubinetti, David R. Slochower, Dongbo Hu, Venkat S. Malladi, Casey S. Greene, Anthony Gitter<br>
 _Manubot Preprint_ (2019) <https://greenelab.github.io/meta-review/>
 
+## Installation
+
```
dhimmel: Let's start off this section with something like the following:

```markdown
If you are using the `manubot` python package as part of a manuscript repository, installation of this package is handled though the Rootstock's [environment specification](https://github.com/manubot/rootstock/blob/master/build/environment.yml).
For other use cases, this package can be installed via pip as follows:
```

agitter: This language looks good to me.  We can capitalize `Python`.

dhimmel:
@agitter I addressed the review comments in https://github.com/manubot/manubot/pull/118/commits/0dda5825ab824190f73f7890fe9bff743eb7f0e7. If looks good to you, I will merge.

PR #119, opened 2019-06-25
Update citation to PLOS CB

cgreene:
Not sure if we want to leave a link to the version of the article produced by manubot.


dhimmel:
Thanks @cgreene for the update. I had also prepared this change with some other README updates but then the internet at Penn wen't down. I will merge your PR first and then rebase mine with the additional changes.

cgreene:
Yours is better than this. Feel free to close this one 😁

Issue #125, opened 2019-07-05
Strip byte order mark


agitter:
In https://github.com/manubot/rootstock/issues/247 a leading byte order mark prevented a section header from rendering correctly.  Should Manubot strip this from the source markdown in
https://github.com/manubot/manubot/blob/025b7c78e96d37cb81388bf739411dfe82969780/manubot/process/manuscript.py#L37

agitter:
This Stack Overview [answer](https://stackoverflow.com/a/45099125) gives two options:
- `lstrip` the '\ufeff' if present
- open with the `utf-8-sig` encoding

We can specifying an encoding here
https://github.com/manubot/manubot/blob/025b7c78e96d37cb81388bf739411dfe82969780/manubot/process/manuscript.py#L35

Per the [`open` docs](https://docs.python.org/3/library/functions.html#open)
> The default encoding is platform dependent (whatever `locale.getpreferredencoding()` returns), but any text encoding supported by Python can be used.

The [`utf_8_sig` encoding](https://docs.python.org/3/library/codecs.html#module-encodings.utf_8_sig) allows an optional byte order mark.  Could we ever read Markdown files that are not UTF-8 encoded?

dhimmel:
From [pandoc's manual](https://pandoc.org/MANUAL.html#character-encoding):

> Pandoc uses the UTF-8 character encoding for both input and output. If your local character encoding is not UTF-8, you should pipe input and output through [`iconv`](http://www.gnu.org/software/libiconv/)

I believe Pandoc removes the BOM at the start of a file ([haskell](https://github.com/jgm/pandoc/blob/f3080c0c22470e7ecccbef86c1cd0b1339a6de9b/src/Text/Pandoc/UTF8.hs#L101-L109), https://github.com/jgm/pandoc/issues/743). However, Pandoc does not output BOMs as per https://github.com/jgm/pandoc/issues/3312.

Here is a script to check whether pandoc removes a BOM:

```bash
printf '\xEF\xBB\xBFtext\n' > pandoc-input.md
pandoc --output=pandoc-output.md pandoc-input.md
file pandoc-input.md pandoc-output.md
```

With pandoc 2.7.2, I got the following output:

```
pandoc-input.md:  UTF-8 Unicode (with BOM) text
pandoc-output.md: ASCII text
```

[...]

PR #127, opened 2019-07-08
Specify encodings: read utf-8-sig, write utf-8

dhimmel:
Closes https://github.com/manubot/manubot/issues/125


dhimmel:
https://github.com/manubot/manubot/pull/127/commits/5ac5fe17f1c590d17cdc9c951e909fb2f137d85b specifies the text encoding for read/write statements. Reading text files uses `utf-8-sig` (to strip BOMs if present). Writing text files uses `utf-8` for compatibility with pandoc and to keep operations consistent across files / platforms. 

https://github.com/manubot/manubot/pull/127/commits/5ac5fe17f1c590d17cdc9c951e909fb2f137d85b does not change the output of `manubot cite` to stdout, which should still use the default `sys.stdout` encoding. Not sure what the implications of changing this would be.

I'll look into whether it makes sense to configure the encoding globally instead, although not sure we can specify a different encoding for reading & writing.

dhimmel:
Some references:

- https://docs.python.org/3/library/locale.html#locale.getdefaultlocale
- https://docs.python.org/3/library/locale.html#locale.getpreferredencoding
- https://docs.python.org/3/using/cmdline.html#envvar-PYTHONIOENCODING
- https://docs.python.org/3/using/cmdline.html#envvar-PYTHONUTF8
- https://docs.python.org/3/using/cmdline.html#envvar-PYTHONCOERCECLOCALE

manubot/process/bibliography.py
```diff
@@ -24,12 +24,12 @@ def load_bibliography(path):
...
         if path.suffix == '.json':
             use_pandoc_citeproc = False
-            with path.open() as read_file:
+            with path.open(encoding='utf-8-sig') as read_file:
                 csl_items = json.load(read_file)
         if path.suffix == '.yaml':
             use_pandoc_citeproc = False
             import yaml
-            with path.open() as read_file:
+            with path.open(encoding='utf-8-sig') as read_file:
```
dhimmel: Possible that `utf-8` alone would be sufficient because I believe the BOM is considered whitespace. Actually, possibly not according to https://stackoverflow.com/a/13156715/4651668.

agitter:
@dhimmel is this ready for review?

dhimmel:
> is this ready for review?

Yes. I was hoping to find a good way to set the encoding globally, but wasn't able to figure out anything satisfactory (at least without more time that is probably not worth it ATM).

dhimmel:
> What is the new file `process/tests/manuscripts/citation-rendering/style.csl` testing?

Accidental tracking. RMed

So I think I've updated the PR such that we always read in `utf-8-sig` and write `utf-8`. That seems like a nice consistent guideline for this project. Now, we haven't applied this to our stderr output (like logging) and our stdout (like `manubot cite`). Perhaps we should in the future. Alternatively users with Python 3.7+ can use the environment variable `PYTHONUTF8=1`.

Issue #120, opened 2019-06-25
Support citation key aliases


nichtich:
I've build a similar workflow based on Pandoc, make, [citation-js](https://citation.js.org/), the command line script [wcite](https://www.npmjs.com/package/wcite) and its Pandoc filter [pwcite](https://www.npmjs.com/package/wcite#filter-pwcite). [p]wcite supports citation key aliases in the special Pandoc metadata field `citekeys`. Here is a minimal example document:

~~~
---
citekeys:
  Vrand04: Q18507561
...

Wikidata is a collaborative knowledge base [@Vrand04].
~~~

In manubot Markdown this would need to be written as:

~~~
Wikidata is a collaborative knowledge base [@wikidata:Q18507561].
~~~

Apart from the prefix (this is another issue), manubot should also be able to process this document:

~~~
---
citekeys:
  Vrand04: wikidata:Q18507561
...

Wikidata is a collaborative knowledge base [@Vrand04].
~~~

This would need to be documented at https://github.com/manubot/rootstock/blob/master/USAGE.md#citations

dhimmel:
Thanks @nichtich for the suggestion. Here's where we're at regarding citation keys.

We currently have a similar concept called citation tags. Currently, manuscripts can define a tabular mapping of citation keys (tags) to the corresponding standard_id. See for example [`citation-tags.tsv`](https://github.com/greenelab/meta-review/blob/master/content/citation-tags.tsv) for the Manubot software paper.

So the functionality we don't currently have is the ability to define citation keys/tags in a YAML frontmatter. However, this feature is planned as part of https://github.com/manubot/manubot/pull/99, which creates a pandoc filter for this package's citation processing. This PR would enable your final example (or something close to it).

There is a separate question of whether we should enable embedded-YAML citation tags for users that are not using the Pandoc filter. There is also a question of whether we should make the pandoc filter the only supported workflow in the future. Currently the issue is that Pandoc's syntax is too restrictive for many standard identifiers (especially URLs and ugly DOIs). Citation tags/keys are a workaround for forbidden characters in a citation identifier, but that requires users to use them (and likely first have processing of a citation fail).

Happy for any advice on how you think we should proceed? I see the proposed pandoc filter as a way that you could most directly use Manubot in your current workflow to enable a wider range of citation sources.

dhimmel:
Another consideration your issue brings up is whether we should use the term "key" rather than "tag". Seems like citekey is a familiar term to LaTeX users?

[...]

Issue #114, opened 2019-06-03
Well defined transformation pipeline for citation identifiers


dhimmel:
We have several types of citation identifiers that potentially required different processing pipelines:

1. citations extracted from a manuscript
2. citations extracted from a manual references file. This category could be subdivided based on what CSL JSON field the citation identifier was extracted from
3. citations passed to the `manubot cite` command

We should think about which processing steps we want to apply in each case. Ideally, we could have modular functions that we pipe together to make it clear and straightforward what transformations will be applied.

Furthermore, we should consider a common nomenclature to use for identifiers based on which transformation steps have been applied. https://github.com/manubot/manubot/pull/113 improves the nomenclature but more is possible.

epogrebnyak:
I got a question: why are you enforcing varieties of identifiers at terminology/notation level and not data structures, eg @dataclass? Terminology convention is a bit hard to follow unless deeply into codebase, while data classes quite unequivocal. 

> Ideally, we could have modular functions

With functions type signatures/hints even better. 

From #129:

> It looks like the major terminology changes are:
> 
>     * `citeproc` to `csl_item`

This is a bibliograhic information in a dict/JSON associated with identifier, right?
 
>     * `citation` and `citation_identifier` to `citekey`

Is this a string starting with `@doi:10/blah` or `doi:10/blah` or `doi, 10/blah` or all of them?

>     * `standard_id` and `standard_citation` to `standard_citekey` (internally)
> 
>     * `short_id` to `short_citekey`
> 

This is point where external reader gets lost. ) Maybe some examples in the docs, or dataclasses even better.

epogrebnyak:
As a follow-up on dataclasses, I think `munubot cite` dataflow can be represented as following sequence of transformations, starting some text string as source (either markdown with `[@citekey]` or one per line citekey text file) and resulting in bibliography rendered by pandoc-cite as text string again. I omit the format (json, palin, etc) for simplicity now.
 
```
text: str -> [CiteKey] -> [Citation] -> [CiteItem] -> csl_style: str -> references: str
```

The main types involved, imo, are `CiteKey`, a tuple of source and indentifier, the `Citation`, parent of `DOI`, `ISBN` and others and ultimately `CiteItem`, a dictionary that is compasable to a CSL JSON.

```
from dataclasses import dataclass

@dataclass
class CiteKey:
    source: str
    identifier: str
    
@dataclass
class Citation:
    identifier: str

    def normalise(self):
        return self 

    def retrieve(self):
[...]

PR #129, opened 2019-07-12
Adopt "citekeys" terminology

dhimmel:
This PR is a work-in-progress to determine the feasibility of adopting the citekeys terminology discussed in https://github.com/manubot/manubot/issues/120#issuecomment-507425730.

While we did recently change terminology in https://github.com/manubot/manubot/pull/113, I think it makes sense to settle on the best terminology sooner rather than later. One big advantage of "citation keys" is that it is widely used and perhaps a bit better understood than "citation ID".

Still lot's TBD, this PR just scratches the surface.


agitter:
I didn't comment in #120, but I'm supportive of switching to more standard pandoc terminology.  I haven't personally explored existing pandoc filters or examples to confirm what that terminology is, but I trust the discussion in #120.

manubot/cite/util.py
```diff
@@ -218,101 +218,106 @@ def is_valid_citation(
...
     if inspection:
-        logging.error(f'invalid {source} citation: {citation}\n{inspection}')
+        logging.error(f'invalid {source} citekey: {citekey}\n{inspection}')
         return False
 
     return True
 
 
-def get_citation_short_id(standard_id):
+def shorten_citekey(citekey):
```
agitter: Would it be more explicit to rename `citekey` as `standard_citekey`?

manubot/cite/util.py
```diff
@@ -218,101 +218,106 @@ def is_valid_citation(
...
     from manubot.cite.citeproc import (
-        citeproc_passthrough,
+        csl_item_passthrough,
         append_to_csl_item_note,
     )
 
     note_text = f'This CSL JSON Item was automatically generated by Manubot v{manubot_version} using citation-by-identifier.'
     note_dict = {
-        'standard_id': citation,
+        'standard_id': citekey,
```
agitter: Are we intentionally keeping the `standard_id` terminology here for now?  It appears to be intended that way per the `csl_item_set_standard_id` docs.

manubot/cite/util.py
```diff
@@ -218,101 +218,106 @@ def is_valid_citation(
...
-            return prefix + citation[len(prefix):]
-    return f'raw:{citation}'
+        if citekey.startswith(prefix):
+            return citekey
+        if citekey.lower().startswith(prefix):
+            return prefix + citekey[len(prefix):]
+    return f'raw:{citekey}'
 
 
 def csl_item_set_standard_id(csl_item):
```
agitter: Keeping `standard_id` here instead of `standard_citekey`?  It looks intentional per the documentation below.

manubot/process/util.py
```diff
@@ -22,43 +21,43 @@
...
+    is_valid_citekey,
+    standardize_citekey,
 )
 
 
-def check_collisions(citation_df):
+def check_collisions(citekeys_df):
     """
-    Check for short_id hash collisions
+    Check for short_key hash collisions
```
agitter: This was called `short_citekey` elsewhere

manubot/process/util.py
```diff
@@ -22,43 +21,43 @@
...
     """
-    collision_df = citation_df[['standard_id', 'short_id']].drop_duplicates()
-    collision_df = collision_df[collision_df.short_id.duplicated(keep=False)]
+    collision_df = citekeys_df[['standard_key', 'short_key']].drop_duplicates()
```
agitter: This was called `standard_citekey` elsewhere

manubot/process/util.py
```diff
@@ -22,43 +21,43 @@
...
-def check_multiple_citation_strings(citation_df):
+def check_multiple_citation_strings(citekeys_df):
     """
     Identify different citation strings referring the the same reference.
     """
     message = textwrap.dedent(f'''\
-    {len(citation_df)} unique citations strings extracted from text
-    {citation_df.standard_id.nunique()} unique standard citations\
+    {len(citekeys_df)} unique citations strings extracted from text
+    {citekeys_df.standard_key.nunique()} unique standard citations\
```
agitter: `standard_citekey` instead here and below?

manubot/process/util.py
```diff
@@ -202,17 +202,17 @@ def get_metadata_and_variables(args):
...
-def get_citation_df(args, text):
+def get_citekeys_df(args, text):
     """
-    Generate citation_df and save it to 'citations.tsv'.
-    citation_df is a pandas.DataFrame with the following columns:
-    - manuscript_id: citation ids extracted from the manuscript content files.
-    - detagged_id: manuscript_id but with tag citations dereferenced
-    - standard_id: detagged_id standardized
-    - short_id: standard_id hashed to create a short base
+    Generate citekeys_df and save it to 'citations.tsv'.
```
agitter: The terminology here seems important.  Do we want to use modifiers of `key` or `citekey`?  The code elsewhere calls some of these entities `*_citekey`.

dhimmel: See docstring for `get_citekeys_df`:

```
    Generate citekeys_df and save it to 'citations.tsv'.
    citekeys_df is a pandas.DataFrame with the following columns:
    - manuscript_key: citation keys extracted from the manuscript content files.
    - detagged_key: manuscript_key but with tag citekeys dereferenced
    - standard_key: detagged_key standardized
    - short_key: standard_key hashed to create a shortened citekey
```

Because these are column names within a "citations" file, I thought it'd be okay to just use a `_key` suffix. However, I think you are right that it'd be a bit more explicit to use `_citekey` as the suffix.

dhimmel: Done in https://github.com/manubot/manubot/pull/129/commits/cf760acc177e0638ec11379af3ff18474c3c3676

manubot/cite/util.py
dhimmel: Done in https://github.com/manubot/manubot/pull/129/commits/cf760acc177e0638ec11379af3ff18474c3c3676

manubot/cite/util.py
dhimmel: That is the current implementation. It has two primary motivations:

1. since CSL using the term `id` for the citekey, it fits the CSL JSON vernacular.
2. easiest for backwards compatibility since we may already have manual-references that are using the `standard_id` in notes.

We could change it at a later or now if you think it's preferable to use `standard_citekey`, `original_standard_citekey`, and `original_citekey` in the notes.

dhimmel:
> It looks like the major terminology changes are:

Yes, will use your descriptions in the squashed commit message (;

> Those are fine with me. They aren't consistently changed everywhere, however, especially in get_citekeys_df.

Fixed.

> Should we document what specifically is meant by citekey, short_citekey, etc.?

Currently, this is done in the `get_citekeys_df` docs. Soon I'd like to extract docs from docstrings... so let's wait till then to consider whether we want a glossary and where.

manubot/cite/util.py
agitter: Because CSL uses the `id` term and the rest of Manubot now uses "citekeys", we'll have to convert between these terms somewhere.  The docstring does explain the correspondence
> The standard_id is extracted from a "standard_citation" field, the "note" field, or the "id" field.

> Note that the Manubot software generally refers to the "id" of a CSL Item as a citekey.
 However, in this context, we use "id" rather than "citekey" for consistency with CSL's "id" field.

I'm undecided about what is least disruptive.  Currently, there are two sets of terms describing the progression from original ids/citekeys to standardized ids/citekeys in `get_citekeys_df` and `csl_item_set_standard_id`.  When reading the CSL JSON, we can have `original_standard_id` or `original_id`.  When reading the manuscript, we have `manuscript_citekey`.  Should these be harmonized?

I also forgot when Manubot would encounter CSL JSON with the `standard_citation` field set.  Was that written by older versions of Manubot?  Or part of the original process for manual references?

dhimmel: > When reading the CSL JSON, we can have original_standard_id or original_id. When reading the manuscript, we have manuscript_citekey. Should these be harmonized?

`original_id` and `original_standard_id` refer to the citekey workflow of a reference whereas `manuscript_citekey` refers to the citekey workflow of a citation. In https://github.com/manubot/manubot/issues/114, I'd like to come to a more harmonized nomenclature and processing workflow. But in the meantime, I think we should proceed with this PR without changing `original_standard_id` and `original_id`. I am learning towards changing them, but not right now and not without a little more thought.

> I also forgot when Manubot would encounter CSL JSON with the standard_citation field set. Was that written by older versions of Manubot? Or part of the original process for manual references?

Previously, manual references would set `standard_citation`, but now the recommendation is to use `id`.

PR #136, opened 2019-09-11
Add GitHub workflow for tests

dhimmel:
This PR is meant to evaluate GitHub actions for testing.

Still need to figure out how to enable pandoc


dhimmel:
Turns out @manubot is not part of the Actions beta release. I requested access, but this may take some time.

Ah but the actions are running for the @dhimmel account https://github.com/dhimmel/manubot/runs/219086993

dhimmel:
I can't figure out how to stop the following error on Linux GitHub builds:

```
pandoc-citeproc: Could not find locale data for C-UTF
```

dhimmel:
From discussion in https://github.com/r-lib/ghactions/issues/262#issuecomment-503156838, it seems that we won't be able to enable pull requests to trigger actions at https://github.com/manubot/manubot that initiate from forks. Instead, the workflows run on the fork and don't show up on the PR, which seems counter-intuitive.

dhimmel:
Okay tests finally passing in https://github.com/dhimmel/manubot/runs/373734247 due to upgrading pandoc.

I guess we may have to merge this PR if we want to have actions on this repo. It's still unclear to me where the action will run post merge... on `manubot/manubot`, on the fork, or on both.

dhimmel:
@agitter I am going to merge to move forward with trying out github actions. Can address any feedback in subsequent PRs.

dhimmel:
I followed up with https://github.com/manubot/manubot/commit/8b6e17724b6257713755a78227248cebc75bddec.

Hopefully this goes away in the future and we can re-enable 3.8.

Now it appears github actions is working properly on pull requests and is running the actions under this repo's scope.

agitter:
I'm doing a post-merge review, and this workflow makes sense.  The workflows seem more complex to me at first because there are so many pre-defined actions and many types of events that can trigger a workflow.

Are the actions themselves versioned releases?  Could you replace `actions/checkout@v1` with [v2 of the checkout action](https://github.com/marketplace/actions/checkout) if you wanted to?

dhimmel:
> Are the actions themselves versioned releases

Yes. We should upgrade to `actions/checkout@v2` now that its available if that doens't break anything

PR #139, opened 2019-09-20
separate api functions and helper functions (in tests)

epogrebnyak:
`util.py` is overloaded both with auxilliary functions and api-level functions in one file which makes it hard  to track transformation and make further changes. Separating tests is a step in preparing `util.py` split.


dhimmel:
Currently, we have a single test file corresponding to each submodule, so it's obvious where tests for a given submodule should go. This PR breask this one-to-one correspondence. However, perhaps it's a special case because `test_api` refers to tests of functions imported in `manubot.cite.__init__.py`. @epogrebnyak is that your thinking?

manubot/cite/tests/test_api.py
```diff
@@ -0,0 +1,149 @@
+import copy
```
dhimmel: Is this import needed?

manubot/cite/tests/test_api.py
```diff
@@ -0,0 +1,149 @@
...
+
+import pytest
+
+from manubot.cite.util import (
+    citekey_to_csl_item,
+    standardize_citekey,
+)
```
dhimmel: If we are interested in testing these functions existence at the `manubot.cite` API level, should we switch import to `from manubot.cite import (`?

epogrebnyak:
> Currently, we have a single test file corresponding to each submodule, 

The idea is to prepare split of util.py into two modules, `cite_api.py` with two higher level functions and the rest of `util.py`, the helper functions. Please see subsequent commits.

epogrebnyak:
Moving things one step ahead - there are two files in this PR in place of `util.py` - they are `citekey.py` and `csl_dict.py` with corresponds well with functions scope in these files - one is changing the citekeys, and another changing CSL dictionaries. I kept two test files for `citekey.py` - one with API-level fucntions and another with the rest of functions, with appropriate docstrings.

I did not think of testing the higher level fucntions as `from manubot.cite import (...` maybe it is  indeed extra clarity, let me try integrate that into a test. I'm hesitant if the file should be `test_citekey_api.py` or just `test_api.py`.

epogrebnyak:
Need some guidance about the use `manubot.cite` functions in other parts of manubot. 

Do you want to keep these imports pointing to individual functions in `manubot.cite.something`? 

Or `append_to_csl_item_note`,  `csl_item_passthrough`,  `csl_item_set_standard_id`  and `shorten_citekey` should be raised to `manubot.cite.__init__.py`? Same questions about `citekey_pattern`, `is_valid_citekey`.

Or point to individual functions and see what it becomes after dataclass refactoring and revisit this issue?

`bibliography.py:`

```
from manubot.cite.citeproc import (
    append_to_csl_item_note,
    csl_item_passthrough,
)
from manubot.cite.util import (
    csl_item_set_standard_id,
    shorten_citekey,
) 
```

`manuscript.py`
```
from manubot.cite.util import (
    citekey_pattern,
    is_valid_citekey,
)
```

manubot/cite/tests/test_api.py
epogrebnyak: Nope, deleted.

epogrebnyak:
Another reference to cit/util.py:

https://github.com/manubot/manubot/blob/71d4d3902a343a99c10d59edd6eb77075f9447de/manubot/process/util.py#L29-L34

epogrebnyak:
Remaining error is due to (seemingly) unrelated bevaiour of zotero / translation server:

```
get_url_csl_item_zotero(url)
Out[19]: 
{'id': '85WSWJH7',
 'type': 'webpage',
 'title': 'pandas-dev/pandas',
 'container-title': 'GitHub',
 'abstract': 'Flexible and powerful data analysis / manipulation library for Python, providing labeled data structures similar to R data.frame objects, statistical functions, and much more - pandas-dev/pandas',
 'URL': 'https://github.com/pandas-dev/pandas',
 'language': 'en',
 'accessed': {'date-parts': [['2019', 9, 22]]}}
```

in tests:
```
csl_item['title'].startswith('Flexible and powerful data analysis')
```

@dhimmel - do you know why this may be happening?

epogrebnyak:
Ok, looks the test for `get_url_csl_item_zotero()` was a temporary problem, now the tests pass, this change can be merged.

manubot/cite/tests/test_url.py
```diff
@@ -48,7 +50,7 @@ def test_get_url_csl_item_zotero_github():
       --data 'https://github.com/pandas-dev/pandas/tree/d5e5bf761092c59eeb9b8750f05f2bc29fb45927' \
       'https://translate.manubot.org/web'
     ```
-    """
+    """
```
dhimmel: trailing space

manubot/cite/citekey.py
```diff
@@ -1,38 +1,7 @@
 import functools
 import logging
-import re
```
dhimmel: This file would be a lot easier to review if the functions were not reordered. Any reason we can't keep the existing order for now?

manubot/cite/tests/test_csl_dict.py
```diff
@@ -0,0 +1,70 @@
+import copy
+import pytest
+
+from manubot.cite.csl_dict import (
+    csl_item_set_standard_id)
```
dhimmel: I think we should call this module `csl_item` rather than `csl_dict`, in line with the current terminology in functions. The "item" terminology is used [by the CSL docs](http://docs.citationstyles.org/en/stable/primer.html#item-metadata):

> Next up are the bibliographic details of the items you wish to cite: the item metadata.
> 
> For example, the bibliographic entry for a journal article may show the names of the authors, the year in which the article was published, the article title, the journal title, the volume and issue in which the article appeared, the page numbers of the article, and the article’s Digital Object Identifier (DOI). All these details help the reader identify and find the referenced work.
> 
> Reference managers make it easy to create a library of items. While many reference managers have their own way of storing item metadata, most support common bibliographic exchange formats such as BibTeX and RIS. The citeproc-js CSL processor introduced a JSON-based format for storing item metadata in a way citeproc-js could understand. Several other CSL processors have since adopted this “CSL JSON” format (also known as “citeproc JSON”).

The terminology we've adopted is `csl_data` for a list of `csl_item` dicts, and `csl_json` for `csl_data` that is JSON-serialized.

Note that Citation Styles Language contains other aspects that metadata storage. It's also a format for specifying a given reference style. Therefore, I could imagine `csl_dict` could be mistaken as somehow referring to a data structure related to the XML styles. Not that item is a very descriptive word, but at least it is what CSL docs seem to use.

manubot/cite/tests/test_url.py
```diff
@@ -38,6 +38,8 @@ def test_get_url_csl_item_zotero_manubot():
     assert [int(x) for x in csl_item['issued']['date-parts'][0]] == [2018, 12, 18]
 
 
+# Note: this test had a temporary failure, maybe due to performance of 
+#       transaltion server.
```
dhimmel: Put this in the docstring so it shows up when the test fails. Typo in `transaltion`.

FYI this is our most common build failure. It seems that sometimes translation-server returns a different title for the same URL. A real mystery. Probably should inquire upstream at some point or change the test. Anyways, this note is helpful.

manubot/cite/citekey.py
```diff
@@ -71,6 +40,68 @@ def standardize_citekey(citekey, warn_if_changed=False):
...
+    short_citekey = shorten_citekey(citekey)
+    csl_item = csl_item_passthrough(csl_item, set_id=short_citekey, prune=prune)
+
+    return csl_item
+
+import functools
+import logging
+import re
+
+# FIXME: this goes away with new classes
```
dhimmel: I'd prefer to remove the `FIXME` statements that refer the proposed class architecture. Helpful to keep a personal record, but best to keep it out of the master code, since this is currently a proposal with an unknown outcome and implementation timeframe that is not required for the current code to function. Let's use `TODO`/`FIXME` when there are actual problems with the code that we are aware of but do not want to currently address.

dhimmel:
>  I kept two test files for citekey.py - one with API-level fucntions and another with the rest of functions, with appropriate docstrings.

Perhaps we should keep all tests for `citekey.py` in a file named `test_citekey.py` for now. I think it's been helpful to have one-to-one correspondence of test files to modules. It's not clear to me what we gain currently by splitting the tests for `citekey.py` into multiple files or keeping a single file but with the name `test_citekey_api.py`.

manubot/cite/tests/test_url.py
dhimmel: Actually what is we just modified the test to accept a title of `pandas-dev/pandas` as well as a title of "Flexible and powerful data analysis...". Then we could add a `TODO` comment to investigate further why tranlsation-server returns different titles intermittently while avoiding our tests from failing.

dhimmel:
@epogrebnyak just checking in since I'm excited to get this PR merged after the comments are addressed.

epogrebnyak:
Thanks for your encouragement, @dhimmel - I almost sorted out the changes
in my mind, will be putting into commits shortly.

ср, 2 окт. 2019 г., 12:32 Daniel Himmelstein <notifications@github.com>:

manubot/cite/citekey.py
epogrebnyak: Ughh.

epogrebnyak: Removed the comments to keep the master cleaner, but the idea in between the lines is the following: a combination of hardcoded tests that fails on forks (#140), cross-referencing to submodule (#145) are heavy toll on modularity and speed of developement. Hope I can propose some changes that make code more maintainable.

epogrebnyak:
>  It's not clear to me what we gain currently by splitting the tests for `citekey.py` into multiple files

We are trying to gradually sort #145, namely multiple entry points to `cite` submodule and isolated code that is not linked to functionality of `manubot cite` command.

epogrebnyak:
@dhimmel - please let me know on what might need to change in current round of review

.gitignore
```diff
@@ -22,3 +22,4 @@ Thumbs.db
 
 ## Text Editors
 .vscode
+.idea/
```
dhimmel:
```suggestion
.idea/

```

manubot/cite/tests/test_citekey_api.py
```diff
@@ -1,56 +1,13 @@
-import copy
+"""Tests API-level functions in manubot.cite. Both fucntions are found in citekey.py"""
```
dhimmel:
```suggestion
"""Tests API-level functions in manubot.cite. Both functions are found in citekey.py"""
```

manubot/cite/tests/test_url.py
```diff
@@ -48,8 +49,19 @@ def test_get_url_csl_item_zotero_github():
...
+
+    Note: this test may have temporary failures, due to performance of
+          translation-server. It seems that sometimes translation-server 
+          returns a different title for the same URL. A real mystery. 
+          
+    See also:
+        https://github.com/manubot/manubot/pull/139#discussion_r328703233 
+
+    Proposed action: 
+        Probably should inquire upstream or change the test.
```
dhimmel: trailing whitespace on these lines

manubot/cite/citekey.py
dhimmel: The import recording here is good, since the diff is simple and we try to keep import alphabetical.

But is the reordering below of whole functions necessary? It makes it harder to review, so unless the new order is highly preferable I think it makes sense to keep the function definition order from manubot/cite/util.py since the diff detects this as the moved file. Would help tremendously with review. I've run `git diff --color-moved=dimmed-zebra master` locally but there it's still a challenge to mentally parse.

manubot/tests/test_imports.py
```diff
@@ -1,10 +1,14 @@
...
-    import manubot.cite.util
     import manubot.cite.arxiv
     import manubot.cite.doi
     import manubot.cite.pubmed
     import manubot.cite.url
     import manubot.process.manuscript
-    assert isinstance(manubot.cite.util.citeproc_retrievers, dict)
+
+# FIXME: citeproc_retrievers are assumed to go away when moving 
+#        functionality to Handle dataclass
```
dhimmel: remove fixme

dhimmel:
Here's the diff I generated to not color relocated lines: [diff.pdf](https://github.com/manubot/manubot/files/3703288/diff.pdf). Used the following code:

```shell
git diff --color --color-moved=dimmed-zebra master | aha > diff.html
wkhtmltopdf diff.html diff.pdf
```

Issue #140, opened 2019-09-22
test_ci.py makes travis checks on a fork to fail


epogrebnyak:
I use manubot repo fork to test code changes, and Travis builds between branches of a fork indicate test success or failure. `test_ci.py` is hardcoded to work on manubot/manubot repo only:

https://github.com/manubot/manubot/blob/307b339a4e600406345d388e20e74cc833cc43c5/manubot/process/tests/test_ci.py#L12-L30

Any ideas on how this can be changed?

dhimmel:
What if we add a second skipif that will skip the test unless the travis environment variable `TRAVIS_REPO_SLUG` equals `manubot/manubot`?

This would weaken the test slightly, for example causing it to be skipped if Travis CI no longer set `TRAVIS_REPO_SLUG`.

Perhaps another option would be to use pytest [custom markers](http://doc.pytest.org/en/latest/example/markers.html). For example, you could decorate tests with `@pytest.mark.forksfail` and then execute pytest with the `-m "not forksfail"` option. But then you'd have to modify `.travis-ci.yml` for your fork... unless we conditionally specified the option in the travis pytest command.

epogrebnyak:
I think we can weaken the tests, eg `assert isinstance(info['repo_slug'], str) and info['repo_slug']`. 

But before that - there must have been a reason you wanted to have checked for `info['repo_slug'] == 'manubot/manubot'`. Is this important for the pdf build process?

dhimmel:
> there must have been a reason you wanted to have checked for `info['repo_slug'] == 'manubot/manubot'`

We wanted to make sure that `get_continuous_integration_parameters` properly detects the repository account/name on GitHub. The test is already weakened by having the account and name both as `manubot` here, so it's looking for `manubot/manubot` rather than two different values such as `manubot/rootstock`. These values are used to template portions of manuscripts that need to link to the GitHub source code.

> I think we can weaken the tests, eg `assert isinstance(info['repo_slug'], str) and info['repo_slug']`, at least on fallback `TRAVIS_REPO_SLUG` =! `manubot/manubot`.

Sounds good to me. While I would like a stronger test, I understand how detecting these environment variables is a difficult thing to hardcode in tests because they change, and therefore weakening is one solution.

epogrebnyak:
More granuarly, we can assume repo name has to be manubot (which is reasonable for a fork):

 `assert info['repo_name'] == 'manubot'`

[...]

PR #142, opened 2019-09-23
Create this_change_triggers_new_build.txt

epogrebnyak:


epogrebnyak:
This branch is not to merge, it is used as testing before https://github.com/manubot/manubot/issues/140 gets any new status. 

I was trying to see the original behaviour of `get_url_csl_item_zotero(url)` - it works in master branch, but fails after `manubot/cite/util.py` refactoring in https://github.com/manubot/manubot/pull/139.

As local tests on Windows fail for manubot and testing in a fork is constrained by #140, I used this PR to test master branch behaviour.

PR #143, opened 2019-09-23
test_remove_jsonschema_errors() forced to use utf-8

epogrebnyak:
The change is needed to make test suite to run on non-latin OEM encoding on Windows.

### DIFF
--- a/manubot/cite/tests/test_citeproc.py
+++ b/manubot/cite/tests/test_citeproc.py
@@ -14,12 +14,17 @@ csl_instances = [
     x.name for x in directory.glob('csl-json/*-csl')
 ]
 
+
+def load_json(path):
+    return json.loads(path.read_text(encoding='utf-8-sig'))
+
+    
 def test_json_is_readable_on_windows_in_different_oem_encoding():
     name = 'crossref-deep-review-csl'
     path = directory / 'csl-json' / name / 'raw.json'
-    content = path.read_text(encoding='utf-8')
+    content = path.read_text(encoding='utf-8-sig')
     assert content
-    json1 = json.load(path.open('r', encoding='utf-8'))
+    json1 = load_json(path)
     assert json1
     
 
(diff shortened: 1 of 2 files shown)


manubot/pandoc/tests/test_bibliography.py
```diff
@@ -43,7 +43,7 @@ def test_load_bibliography_from_text(path):
     """
     https://zbib.org/c7f95cdef6d6409c92ffde24d519435d
     """
-    text = path.read_text()
+    text = path.read_text(encoding='utf-8')
```
dhimmel: Let's use `utf-8-sig` as per discussion in https://github.com/manubot/manubot/pull/127

manubot/cite/tests/test_citeproc.py
```diff
@@ -28,10 +36,12 @@ def test_remove_jsonschema_errors(name):
...
-    """
-    data_dir = directory / 'csl-json' / name
-    raw = json.loads(data_dir.joinpath('raw.json').read_text())
-    expected = json.loads(data_dir.joinpath('pruned.json').read_text())
+    """    
+    data_dir = directory / 'csl-json' / name 
+    def load_json(filename):
+        return json.load((data_dir / filename).open('r', encoding='utf-8'))
+    raw = load_json('raw.json')
+    expected = load_json('pruned.json')
```
dhimmel: Any reason to change more than just adding `encoding='utf-8-sig'` as an arg to `.read_text`? I don't love defining this function to load a single json when the existing functions support everything needed

epogrebnyak: > existing functions support everything needed
Actually there was no single function, but repeated code - one needs to copy-paste `encoding='utf-8-sig'` into several statements. 

Also `raw = json.loads(data_dir.joinpath('raw.json').read_text())` combines operations at different level of abstraction - it continues to specifiy the location of the file and the read operation. That makes for nested parenthesis, which are hard to read through quickly. Under these considerations I came to be thinking of `load_json(filename)`.

Issue #144, opened 2019-10-08
test_cite_command_render_stdout[--format=jats] fails with pandoc 2.7.3


epogrebnyak:
To be further investigated, maybe utf-8 local file reading issue similar to #143

dhimmel:
Here's the source code
https://github.com/manubot/manubot/blob/df480ae2f9de33569a040de420d6a57924927dfe/manubot/cite/tests/test_cite_command.py#L68-L99

Looks like we need to specify the encoding in `.read_text`. There failures are showing the value of testing on a Windows system whose default encoding is not UTF-8.

epogrebnyak:
Another little mistery - I'm running a command line with `--format jats` and saving the output to jats.txt:

```
manubot cite --render --csl https://github.com/greenelab/manubot-rootstock/raw/e83e51dcd89256403bb787c3d9a46e4ee8d04a9e/build/assets/style.csl arxiv:1806.05726v1 doi:10.7717/peerj.338 pmid:29618526 --format jats > jats.txt
```

then doing file compare with reference file - `references-jats.xml` and the output is slightly different in schema of the files. Ok, but why does this happen on a local Windows machine and is not an error on a remote linux machine?  Any clues?

My local panodc is 2.7.3.

Here is a lengthy file compare, don't know how to highlight it better:

```
λ fc references-jats.xml jats.txt
Сравнение файлов references-jats.xml и JATS.TXT
***** references-jats.xml
<article-meta>
</article-meta>
***** JATS.TXT
<article-meta>
<pub-date pub-type="epub">
<year></year>
</pub-date>
</article-meta>
*****

***** references-jats.xml
  location</article-title><source>PeerJ</source><publisher-name>PeerJ</publisher-name><date><day>03</day><month>04</month><year
>2014</year></date><volume>2</volume><pub-id pub-id-type="doi">10.7717/peerj.338</pub-id><ext-link ext-link-type="pmid" xlink:h
ref="http://www.ncbi.nlm.nih.gov/pubmed/<ext-link ext-link-type="uri" xlink:href="https://www.ncbi.nlm.nih.gov/pubmed/24765572"
>24765572</ext-link>" xlink:type="simple"><ext-link ext-link-type="uri" xlink:href="https://www.ncbi.nlm.nih.gov/pubmed/2476557
2">24765572</ext-link></ext-link><ext-link ext-link-type="uri" xlink:href="https://doi.org/gfrfbm">https://doi.org/gfrfbm</ext-
[...]

Issue #117, opened 2019-06-24
Choose service for docs generation


vincerubinetti:
We should choose a service/product to help us write and generate the docs for all of Manubot. It can be a full service that manages the writing of the docs and also hosts it (like `readthedocs`) or it can just be some kind of tool that generates a static docs website from some simple markdown. 

Here are the most popular ones, it seems:

[readthedocs](https://readthedocs.org/)
[sphinx](http://www.sphinx-doc.org/en/master/)
[gitbook](https://www.gitbook.com/pricing)

We could even just do it "manually" with Jekyll, and it might not be much more difficult than a docs-specific generation tool.

dhimmel:
One consideration will be whether we want the docs for the `manubot` python package also in this location. I think so? If so, we will want a way to automatically generate rendered documentation from the python source code. I believe at least some of the services above support this.

Currently, it seems like we have at least three sources of documentation:

1. Usage documentation from rootstock's USAGE.md
2. Setup documentation from rootstock SETUP.md
3. Documentation for the `manubot` python package

Another consideration is versioning. Currently, we version based on the git commit hashes, but this may no longer be possible or desirable?

epogrebnyak:
Correction: readthedocs  is a rendering service using sphinx. Gitbook is ok for generating for markdown-based sites, bit not documentation from code. We are pretty much constrained to sphinx on python.

dhimmel:
One option I'd like to look into is https://github.com/timothycrosley/portray.

dhimmel:
Docs are now available, auto-generated using portray, at https://manubot.github.io/manubot/. Not perfect, but enough to let us browse the docs and see where we need to improve.

PR #146, opened 2019-10-08
supplementary xml file for --jats renderer (addresses #144)

epogrebnyak:


manubot/cite/tests/test_cite_command.py
```diff
@@ -65,38 +61,75 @@ def test_cite_command_file(tmpdir):
...
-    print(process.stdout)
-    print(process.stderr)
-    assert process.stdout == expected
+class Base_cite_command_render_stdout():
+    @classmethod
+    def expected_output(cls, filename):
+        return (
+            pathlib.Path(__file__).parent
+            .joinpath('cite-command-rendered', filename)
+            .read_text()
```
dhimmel: let's specify `utf-8-sig` as per our convention

manubot/cite/tests/test_cite_command.py
```diff
@@ -65,38 +61,75 @@ def test_cite_command_file(tmpdir):
...
+        assert self.render(['--format', 'markdown']) == \
+            self.expected_output('references-markdown.md')
+
+    def test_html(self):
+        assert self.render(['--format', 'html']) == \
+            self.expected_output('references-html.html')
+
+    def test_jats(self):
+        if pandoc_version >= (2, 7, 3):
+            filename = 'references-jats-2.7.3.xml'
```
dhimmel: Would be best to have no hardcoded version paths in this test file.

Assume the following:

1. We will frequently be updating the Pandoc version used for CI.
2. Everytime we update pandoc, all outputs (not just JATS) have a strong probability of breaking.

Therefore, I think it's best if we have the tests detect the pandoc version and look for the appropriate output. But we want to avoid having to change the test everytime we update the pandoc version and create new outputs. Does that make sense?

epogrebnyak: > have no hardcoded version paths in this test file

> we have the tests detect the pandoc version and look for the appropriate output.

Not sure what you mean - what kind of output the tests would need to detect? It must be a straightforward idea, but I'm not catching it.

dhimmel: Something that works for all formats / versions and doesn't require updating `test_cite_command.py` with special conditions. Something along the lines of:

```python
# Something like this to generate filename
filename = f"references-{format}-{'.'.join(pandoc_version)}.{extension}"
if not filename.exists():
    assert 'CI' not in os.environ
    # warning like "please create {filename} for test to proceed"
    pytest.skip(...)
# Proceed with filename.read_text and test
```

My experience has been that more often than not, pandoc updates do change the output, so we might as well assume they always do and only evaluate if the version is correct. Does that make sense?

I don't think it's a huge deal if the test is skipped locally... but we could issue a warning so the user is aware.

epogrebnyak: Here is a version of the logic that may be compatible with your suggestion as I understand it: 
1. we are sure about pandoc version in on the CI, because we have installed it there and produced the output files with that version
2. producing new output files should not touch the code of tests and be a change of artefacts, not the code
3. because we are never sure of local pandoc setting we can sacrifice the strengths on local tests

There is a cost:
- with `f"references-{format}-{'.'.join(pandoc_version)}.{extension}"` we accrue the complexity of naming (means we might need special tools/scripts to make those files). 

Also on subjectively on implementation:
- ideally, a test should be ambiguious, mixing warnigns and test results sends a noisy signal
- `assert 'CI' not in os.environ` is tempting to write, but it is a shorthand, because assert may be [disabled by `-o`  flag](https://stackoverflow.com/questions/1273211/disable-assertions-in-python)

Suggestion: if we can to make all output file names contain a version, as you proposed, maybe there can be a rather simple rule for a test:
- find output file that matches installed pandoc version
- if found a match, run test on it
   - if success, move on to next test
   - if test failed - fail
- if no version matched - try other output file versions (there is probably just one)

So, if we keep `reference-jats-2.7.3.txt` and the rest of `reference-{format}-2.7.2.txt` files in the repo, both remote and local versions would work. This would also simplify the test lookup.

manubot/cite/tests/test_cite_command.py
epogrebnyak: Maybe need a more global helper funciton - I'll put a note on refactoring proposal in a comment.

manubot/pandoc/util.py
```diff
@@ -43,3 +43,11 @@ def get_pandoc_info():
         stats[f'{command} path'] = path
     logging.info('\n'.join(f'{k}: {v}' for k, v in stats.items()))
     return stats
+
+
+def get_pandoc_version() -> (int, int) :
```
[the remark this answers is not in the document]
epogrebnyak: https://github.com/manubot/manubot/pull/146/commits/c05f12c9a1edab0d6bb9f17f0f5ce276a7fb0e3b

manubot/cite/tests/test_cite_command.py
epogrebnyak: Apparently `utf-8-sig` https://ci.appveyor.com/project/manubot/manubot/builds/28002722/job/vy7a7kbigj665ewx causes problems on Appveyor, not sure why.

epogrebnyak:
```
    def test_no_arg(self):
>       assert self.render([]) == \
            self.expected_output('references-plain.txt')
E       AssertionError: assert '1. Generaliz... PMC5938574\n' == '1. Generaliz... PMC5938574\n'
E         Skipping 318 identical leading characters in diff, use -v to show
E         - peerj.338 · PMID: 24765572 · PMCID: PMC3994616
E         ?           -                 -
E         + peerj.338 � PMID: 24765572 � PMCID: PMC3994616
E           
E           3. Opportunities and obstacles for deep learning in biology and medicine.
E         - Travers Ching, Daniel S Himmelstein, Brett K Beaulieu-Jones, Alexandr A Kalinin, Brian T Do, Gregory P Way, Enrico Ferrero, Paul-Michael Agapow, Michael Zietz, Michael M Hoffman, … Casey S Greene
E         ?                                                                                                                                                                                    ^^^
E         + Travers Ching, Daniel S Himmelstein, Brett K Beaulieu-Jones, Alexandr A Kalinin, Brian T Do, Gregory P Way, Enrico Ferrero, Paul-Michael Agapow, Michael Zietz, Michael M Hoffman, � Casey S Greene
E         ?                                                                                                                                                                                    ^
E           Journal of the Royal Society, Interface (2018-04) https://www.ncbi.nlm.nih.gov/pubmed/29618526
E         - DOI: 10.1098/rsif.2017.0387 · PMID: 29618526 · PMCID: PMC5938574
E         ?                             -                 -
E         + DOI: 10.1098/rsif.2017.0387 � PMID: 29618526 � PMCID: PMC5938574
manubot\cite\tests\test_cite_command.py:106: AssertionError
```

manubot/cite/tests/test_cite_command.py
epogrebnyak: As for `assert`s I've encountered its use inside code outside tests, that's why I pointed to the usage in tests:  

https://github.com/manubot/manubot/blob/cf9564dd4a31701e48a487ecd716b2cea6aca53a/manubot/cite/csl_item.py#L43

One can think of `assert` vs throughing exceptions as matter of style, but I think exceptions are cleaner, based on [this](https://stackoverflow.com/questions/944592/best-practice-for-python-assert) and [this](https://stackoverflow.com/questions/42837054/are-assert-methods-acceptable-in-production), that is why I called `assert` a shorthand, something you write in place of a longer, but more stable expression. 

In your test example the asset was not related to code under test, so it seemed a shorthand too.

manubot/cite/tests/test_cite_command.py
epogrebnyak: Which means system encoding is different on Appv and Travis machines? Strange, but let's stip the encoding with comment.

dhimmel: > Which means system encoding is different on Appv and Travis machines?

Yeah, well AppVeyor is testing on Windows, Travis on Linux.

epogrebnyak: > AppVeyor is testing on Windows

That I did not realise! The screens looked like Powershell though.

manubot/cite/tests/test_cite_command.py
```diff
@@ -65,38 +61,101 @@ def test_cite_command_file(tmpdir):
...
+            pathlib.Path(__file__).parent
+            .joinpath('cite-command-rendered', filename)
+            .read_text()
+        )
+        # Note: We wanted to introduce encoding='utf-8-sig' above
+        # for compatability, but that makes tests fail on Travis for 
+        # Windows default encoding.
+
+    @classmethod
+    def render(self, format_args):
```
dhimmel: decorator is classmethod, but method defition uses `self` not `cls` and code uses neither, such that it probably should be `@staticmethod`?

manubot/cite/tests/test_cite_command.py
```diff
@@ -65,38 +61,101 @@ def test_cite_command_file(tmpdir):
...
+            stderr=subprocess.PIPE,
+            universal_newlines=True,
+        )
+        print(shlex_join(process.args))
+        print(process.stdout)
+        print(process.stderr)
+        return process.stdout
+
+
+@pytest.mark.pandocfail
```
dhimmel: What does this marker mean?

epogrebnyak:
> @epogrebnyak did you decide that you don't want to skip the test if the correct version is not located on a non-CI build?

The tests implementation is for CI builds as the version is known there beforehand. AS for local builds we allow a user to skip the tests is she find a local pandoc version is different from CI versions. To skip the tests a user can run `pytest -v -m "not pandocfail"`. 

This approach is fairly documented in docstring below. Improvements I can think of:

- [ ] change the tests marker (something different that `pandocfail`)
- [ ] exclude exact current version number from a docstring
- [ ] add a comment about local tests into README / dev manual / docs

I decided not to shut the tests based on 'non-CI build' - skipif's make tests logic more complicated as in #148. I suggest keeping a base version of the tests that support the CI (as it has more stable environment) and provide the local user the instructions on skipping tests that may fail lcoally.  

```
"""
    Expecting reference values for test to be at files on path:
    cite-command-rendered/references-{format}-{pandoc_stamp}.{extension}

    Examples:
    - references-html-2.7.2.html  
    - references-plain-2.7.2.txt
    - references-jats-2.7.2.xml
    - references-jats-2.7.3.xml
    - references-markdown-2.7.2.md

    2.7.2 is current pandoc version on CI builds for these builds.

    Filenames must be adjusted accodingly when current pandoc version changes.

    Run tests locally skipping this test suit (makes sense if your local pandoc
    version is different from pandoc version used by Travis and Appveyor):

        pytest -v -m "not pandocfail"
        pytest -k "Test_cite_command_render_stdout" -m "not pandocfail"
        
    See .travis.yml and .appveyor.yml to find out current pandoc version used 
    for testing.
    """
```

manubot/cite/tests/test_cite_command.py
epogrebnyak: Please see comment below

manubot/cite/tests/test_cite_command.py
epogrebnyak: Indeed! `self` is not used here.

epogrebnyak:
Refactoring idea, not todo now: can have a base class for `pandoc` / `pandoc-citeproc` tests that has the skipif's. We use calling the command line in several places, so maybe one parent for these tests makes sense. 

Then


```
@pytest.mark.skipif(
    not shutil.which('pandoc'),
    reason='pandoc installation not found on system'
)
@pytest.mark.skipif(
    not shutil.which('pandoc-citeproc'),
    reason='pandoc-citeproc installation not found on system'
)
class Base_cite_command_render_stdout():
```

becomes 

`class Base_cite_command_render_stdout(Base_pandoc):`

`Base_pandoc` can also have the `run()` staticmethod to reuse this code:

```
        process = subprocess.run(
            args,
            stdout=subprocess.PIPE,
            stderr=subprocess.PIPE,
            universal_newlines=True,
        )
```

epogrebnyak:
Lets see if the tests pass on CI...

manubot/cite/tests/test_cite_command.py
```diff
@@ -65,38 +61,103 @@ def test_cite_command_file(tmpdir):
...
+
+    2.7.2 is current pandoc version on CI builds for these builds.
+
+    Filenames must be adjusted accodingly when current pandoc version changes.
+
+    Run tests locally skipping this test suit (makes sense if your local pandoc
+    version is different from pandoc version used by Travis and Appveyor):
+
+        pytest -v -m "not pandocfail"
+        pytest -k "Test_cite_command_render_stdout" -m "not pandocfail"
```
dhimmel: I'm not sure `pandocfail` is the best term, since this marker isn't applied to all tests that require pandoc. Maybe something like `pandoc-version-dependent`? But also we're not using this anywhere else. So for now would it just be easier for local users to use `pytest -k "not Test_cite_command_render_stdout"` and we'll include that in the docstring?

epogrebnyak: I left a marker - it is safer for local user to depend on a marker, than requiring a user to know exact test name pattern to skip it. We already penalise a local tester with additional command line setups, so let's at least make that additional use sensible. 

Also need dearly a place to store developper manual,  #117. It will be rather hard to dig back to collect dev notes in retrospect.

epogrebnyak:
Must integrate a fixture from #151 once it is in master branch.

dhimmel:
> Must integrate a fixture from #151 once it is in master branch.

Am merging now since this looks sufficient to address https://github.com/manubot/manubot/issues/144. Can make a new PR for these additional integrations.

PR #147, opened 2019-10-09
Introducing dataclasses (not to merge yet)

epogrebnyak:
This branch introduces dataclasses into `manubot.cite` data model. Key classes are `Citekey`, `Handle` (parent of DOI, ISBN and other handles) and `CSL_Item`. These class definitions are in `types.py`.

I introduced just one actual handle class `DOI` in `doi.py` and used it in transformation of citekey in `standardize_citekey()` function. Getting a standardized identifier now looks like:

```python
identifier = CiteKey(citekey).handle().canonic().identifier
```

Next refactoring is to introduce `ISBN` class (very easy), so that there will be just one transformation line instead of two if's in `standardize_citekey()`. I postpone this refactoring until reviewers get familiarised with and approve the solution. 

Main benefits are:
- get rid of case-like branching `if source=...` and associated machinery of retriever function lookups 
- bringing the functionality of working with identifiers closer to where they are (eg make all work with DOI in doi.py, including inspect functions)
- provide methods for uniform chains of transformations from Citekey to CSL_Item (as in example above), see also https://github.com/manubot/manubot/issues/114 on this.

In current implementation things to watch/change are:
1. all new dataclasses are in types.py, to avoid circular references. Respected classes more belong to `citekey.py` and `csl_item.py`. 
2. must allocate existing tests and make new tests is a todo, there are just a few asserts now
3. may clean `CSL_Item` dummy methods in `types.py` until we need them for refactoring functions other than 

This is a trail run of a bit of architecture change, so comments are highly welcome.


epogrebnyak:
`dataclasses` is a package in 3.6 and a part of standard library in 3.7.

Should I just add `dataclasses` library to `setup.py` to make things work for python 3.6? Or maybe there is a tricky way to install it just for python 3.6? 

```
    # Run-time dependencies
    install_requires=[
        'errorhandler',
        'isbnlib',
        'jinja2',
        'jsonref',
        'jsonschema',
        'packaging',
        'pandas',
        'pybase62',
        'pyyaml',
        'ratelimiter',
        'requests-cache',
        'requests',
    ],
```

manubot/cite/doi.py
```diff
@@ -7,6 +7,32 @@
 from manubot.cite.pubmed import get_pubmed_ids_for_doi
 from manubot.util import get_manubot_user_agent
 
+from manubot.cite.types import Handle, CSL_Item
+
+class DOI(Handle):
+    def canonic(self):
+        self.identifier = self.identifier.lower()
```
dhimmel: switch order of lowercasing and expanding so lowercasing is done last

manubot/cite/types.py
```diff
@@ -0,0 +1,158 @@
...
+        """
+        constr = cls._get_constructor(source)
+        return constr(identifer)
+
+
+@dataclass
+class Handle(HandleBase):
+    identifier: str
+
+    # Child classes must redefine canonic() and csl_item()
```
dhimmel: Is using `@abc.abstractmethod` a way to enforce this?

dhimmel:
> dataclasses is a package in 3.6 and a part of standard library in 3.7.

You can use an [Environment Marker](https://www.python.org/dev/peps/pep-0508/#environment-markers) in `install_requires`.

manubot/cite/doi.py
epogrebnyak: cae3e5f

manubot/cite/types.py
epogrebnyak: For `csl_item()` - probably yes, for `canonic()` probably not. `csl_item()` has to change in every child class, while `canonic()` stays the same for all classes except `DOI` and `ISBN`. 
I did not use ABCs before, so unfamiliar with caveats.

epogrebnyak:
Seems to be failing on a single branch in Appveyor when quering a translation server: https://ci.appveyor.com/project/manubot/manubot/builds/28009208/job/ry8wknsss3w3akof

epogrebnyak:
An update on status on this branch:

> - all new dataclasses are in types.py, to avoid circular references. Respected classes more belong to citekey.py and csl_item.py.
- [x] `Citekey` now resides in citekey.py, `CSL_Item` in csl_item.py, `types.py` contains only handle class and renamed to handle.py
- [x] circular references avoided by moving some imports into function-level scope

> must allocate existing tests and make new tests is a todo, there are just a few asserts now
- [x] asserts made tests
- [ ] still can do a bit more of testing of new classes own functionality
- [ ] existing tests not touched

> may clean CSL_Item dummy methods in types.py until we need them for refactoring functions 
- [ ] still a todo

epogrebnyak:
My proposed roadmap is to review this branch and pick things incrementally to submit to master by other PRs:

Round 1:

- [ ] standardize_citekey() can be split into worker function and logger function (38bcd8c)
- [ ] standardize_citekey() is consumer of `Citekey` and `Handle` classes. This fucntion result can be expressed as: `str(CiteKey(citekey).handle().canonic().citekey())`. I would like to hear your opinion if this pipeline makes sense or looks like a mess to you. I'm too familiarised at this point with the proposed dataclasses to judge.

Round 2

- [ ] we can bring csl item transformations closer to CSL_Item class, by putting primitive methods needed to replicate the `csl_item_passthrough` and `append_to_csl_item_note` functions.

Round 3

- [ ] we link individual retriver fucntions as .csl_item() methods (functions stay as they are, the methods call them) and simplify the retriver lookup architecture.

These changes are rather isolated - round 1 does not even require `CSL_Item`, round 2 introduces the `CSL_Item` into existing program flow, but program still does not rely on it retrive the bibliographic information and then round 3 conludes the pipeline. 

I think when doing 1 and 2 we will find interesting behaviours that prepare us better to the switch at round 3. 

There is alos round 4, refactoring utilities used by larger manubot, but not touched by `manibot.cite`.

manubot/cite/csl_item.py
```diff
@@ -1,6 +1,201 @@
...
+csl_item_type_fixer = {
+    'journal-article': 'article-journal',
+    'book-chapter': 'chapter',
+    'posted-content': 'manuscript',
+    'proceedings-article': 'paper-conference',
+    'standard': 'entry',
+    'reference-entry': 'entry',
+}
+
+class CSL_Item(dict):
```
dhimmel: I wonder if this should extend `collections.OrderedDict`? Plain dicts have been ordered since 3.6, but some software still doesn't treat them as such (but does treat OrderedDicts as ordered). I think extending `dict` is fine unless testing determines that some non-determinism or sorting by key gets introduced.

manubot/cite/csl_item.py
```diff
@@ -1,6 +1,201 @@
...
+    #     # Correct invalid CSL item types
+    #     # See https://github.com/CrossRef/rest-api-doc/issues/187
+    #     if 'type' in csl_item:
+    #         csl_item['type'] = csl_item_type_fixer.get(csl_item['type'], csl_item['type'])
+    
+    def fix_type(self):
+        # Correct invalid CSL item types
+        # See https://github.com/CrossRef/rest-api-doc/issues/187
+        try:
+            self['type'] = csl_item_type_fixer.get(self['type'])
```
dhimmel: .get should default to `csl_item['type']` as before otherwise it can set an existing type to None.

dhimmel:
> standardize_citekey ... can be expressed as: `str(CiteKey(citekey).handle().canonic().citekey())`

Do we expect users to be interacting with these functions? If so I think it's a lot of steps. From the users perspective would the following be simpler:

```python
standard_citekey = CiteKey('doi:10/xxx').canonic().__str__()
```

I need to review a bit close to see when new objects are created versus returning the current object modified. And which objects are of which classes.

manubot/cite/csl_item.py
epogrebnyak: I did not think of OrderedDict, but I experimented with UserDict. Unfortunately UserDict had problems qith serialisation, I could not get it to work with json.dump(). 

I used dict because it is used now for csl_item.

manubot/cite/csl_item.py
epogrebnyak: Maybe this is better: 

self['type'] = csl_item_type_fixer[self['type']]

Inside try-except this will fail and pass if there is no type, or type not a key in a fixer dict.

epogrebnyak:
The proposed dataclass methods are designed to work on the data structures inside those classes. '.canonic()' is specific to DOI and ISBN, and defined for Handle class and subclasses.

Whenever you standardise a citekey, you get a handle, apply canonic() to it and then tune the result to the type as string. 

In this approach you have fewer switches (just one, choosing constructor) and recycle the code that converts between classes. 

There can be some sugar added though to make a higher level API, such as Citekey(citekey).standardize() (effectively same function we have now). I experimented with it a bit, but decided we need first to work on lower-level implementation, that csn enable the sugarisation.

epogrebnyak:
Here is an example of sugarisation:

```python
class Citekey:
    def standardize(self):        
        handle = self.handle().canonic()
        return handle.citekey()

assert CiteKey('doi:10.1038/SREP16899').standardize().str() = 'doi:10.1038/srep16899'
```
```

**Comment by epogrebnyak** [comment]
> I need to review a bit close to see when new objects are created versus returning the current object modified. And which objects are of which classes.

One idea about making the proposed API easier - make conversion between classes a property and methods that work on a class instance - regular methods. This way we have less parens in transfomation: 

```
handle = Citeykey('doi:10.1038/SREP16899').handle.canonic() # instead of .handle()
citekey_str = str(handle.citekey) # instead of .citekey()
```

**Comment by epogrebnyak** [comment]
So the whole type-method system for proposed dataclasses can be sketched as: 

```
CiteKey(citekey: str)
    .handle -> Handle  

Handle(identifier: str)
    .citekey   -> CiteKey
    .canonic() -> Handle  
    .csl_item  -> CSL_Item

CSL_Item(incoming_dict)
```

**Comment by epogrebnyak** [comment]
> when new objects are created versus returning the current object modified. 

This is not a big difference for the `CiteKey` and `Handle` classes , as is the instance is just one or two variables. We both change the object (if someone wants inline change) and return it (for chaining transformations).

For `CSL_Item` we modify `self` and return it because it dict has has many fields. But for example `CSL_Item.minimal()` is creating a new instance without altering the existing instance.

**Comment by epogrebnyak** [comment]
Should probably update this branch with master and set checkpoints for dataclass adoption depending on scope of changes.

File: manubot/cite/csl_item.py (hunk 3 above)
Comment (dhimmel) [reply manubot/cite/csl_item.py]: extending dict is correct. I looked into this more and basically `collections.OrderedDict` should be used as little as possible given that in the future it should never be needed. Therefore if we need `collections.OrderedDict` for something to work, we can cast to it when needed but not before.

File: manubot/cite/csl_item.py [hunk 5]
```diff
@@ -28,6 +28,197 @@
...
+    Return the dictionary of key-value pairs encoded in a CSL JSON note.
+    Extracts both forms (line-entry and braced-entry) of key-value pairs from "cheater syntax"
+    https://github.com/Juris-M/citeproc-js-docs/blob/93d7991d42b4a96b74b7281f38e168e365847e40/csl-json/markup.rst#cheater-syntax-for-odd-fields
+    """
+    note = str(note) # type safeguard 
+    line_matches = re.findall(
+        r'^(?P<key>[A-Z]+|[-_a-z]+): *(?P<value>.+?) *$', note, re.MULTILINE)
+    braced_matches = re.findall(
+        r'{:(?P<key>[A-Z]+|[-_a-z]+): *(?P<value>.+?) *}', note)
+    return dict(line_matches + braced_matches)
```
dhimmel: The changes to csl_item have now been merged in other PRs. So can you merge with the latest master and remove these changes from this PR?

dhimmel:
Much of this functionality was added in https://github.com/manubot/manubot/pull/147. Thanks @epogrebnyak for inspiring these changes with your thoughts and implementation here.

PR #148, opened 2019-10-09
Relax repo tests for #140

epogrebnyak:
Grouped tests into several methods, excluding head-to-head checks for repo slug in most of them. 

Several tests run if repo slug is "manubot/manubot". 

Expected behavior: none of online tests is skipped in master repo. 

This test suit passes on a fork.

### DIFF
--- a/manubot/process/tests/test_ci.py
+++ b/manubot/process/tests/test_ci.py
@@ -8,21 +8,44 @@ from ..ci import (
     get_continuous_integration_parameters,
 )
 
+@pytest.fixture
+def info():
+    return get_continuous_integration_parameters()
 
-@pytest.mark.skipif('TRAVIS' not in os.environ, reason='tests environment variables set by Travis builds only')
-def test_get_continuous_integration_parameters_travis():
-    info = get_continuous_integration_parameters()
+
+# TODO: test generic properties of `info` distionary, eg list of keys
+# TODO: make class for get_continuous_integration_parameters
+
+@pytest.mark.skipif('CI' not in os.environ)
+def test_get_continuous_integration_parameters_generic(info):
     assert info is not None
-    assert info['provider'] == 'travis'
-    assert info['repo_slug'] == 'manubot/manubot'
-    assert info['repo_owner'] == 'manubot'
-    assert info['repo_name'] == 'manubot'
+    # assume fork repo name unchanged
+    assert info['repo_slug'].endswith('manubot') 
+    assert info['repo_name'] == 'manubot' 
+    # commis are not empty
     assert info['commit']
     assert info['triggering_commit']
+
+
+# TODO: need similar test for APPVEYOR, what is the slug constant name?
+#       may embed in this test  
+@pytest.mark.skipif('TRAVIS' not in os.environ 
+                     or os.environ['TRAVIS_REPO_SLUG'] != 'manubot/manubot',
+                     reason='this test runs only for ')
+def test_get_continuous_integration_parameters_on_master_repo_on_travis(info):
+    assert info['repo_slug'] == 'manubot/manubot'
+    assert info['repo_owner'] == 'manubot'
+
+@pytest.mark.skipif('TRAVIS' not in os.environ, reason='tests environment variables set by Travis builds only')
+def test_get_continuous_integration_parameters_travis(info):
+    assert info is not None
+    assert info['provider'] == 'travis'
     assert info['build_url'].startswith('https://travis-ci.com/manubot/manubot/builds/')
     assert info['job_url'].startswith('https://travis-ci.com/manubot/manubot/jobs/')
+    
+    # TODO: must be separate test - it depends on owner name
     # test add_manuscript_urls_to_ci_params
-    info_updated = add_manuscript_urls_to_ci_params(info)
+    info_updated = add_manuscript_urls_to_ci_params(info)    
     assert info is info_updated
     assert re.fullmatch(
         pattern=r"https://manubot\.github\.io/manubot/v/[0-9a-f]{40}/",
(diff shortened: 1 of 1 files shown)


manubot/process/tests/test_ci.py
```diff
@@ -9,50 +9,92 @@
...
+        # https://www.appveyor.com/docs/environment-variables/
+        #
+        # TRAVIS_REPO_SLUG: The slug (in form: owner_name/repo_name) of the repository currently being built.
+        # https://docs.travis-ci.com/user/environment-variables/
+        if 'TRAVIS' in os.environ:
+            assert info['repo_slug'] == os.environ['TRAVIS_REPO_SLUG']
+        elif 'APPVEYOR' in os.environ:
+            assert info['repo_slug'] == os.environ['APPVEYOR_PROJECT_SLUG']
+        else:
+            assert False
```
dhimmel: So this will fail if CI is run on a different service than Travis or AppVeyor? My instinct is that it should pass as the test is for whether lookup succeeds on the *supported* CI platforms, not whether it works on every possible platform. Does that make sense?

manubot/process/tests/test_ci.py
```diff
@@ -9,50 +9,92 @@
...
+
+    def test_commits_are_not_empty(self, info):
+        assert info['commit']
+        assert info['triggering_commit']
```
dhimmel: Afraid this will fail on unsupported CI platforms like in #136. Anyway to make this dependent on a supported CI platform having been detected?

manubot/process/tests/test_ci.py
```diff
@@ -9,50 +9,92 @@
...
-    # test add_manuscript_urls_to_ci_params
-    info_updated = add_manuscript_urls_to_ci_params(info)
-    assert info is info_updated
-    assert re.fullmatch(
-        pattern=r"https://manubot\.github\.io/manubot/v/[0-9a-f]{40}/",
-        string=info['manuscript_url'],
-    )
+@pytest.fixture
+def info():
+    return get_continuous_integration_parameters()
```
dhimmel: Is the fixture rerun for every class method where its invoked or is the same info output reused? This is a general question about how fixtures work.

manubot/process/tests/test_ci.py
epogrebnyak: This is similar to code already used, or I'm missing something?

manubot/process/tests/test_ci.py
epogrebnyak: It should be a rerun, I think, but it is not expensive - should not be slowing the tests too much.

epogrebnyak:
Oh, it passes on a fork, but misses on 'manubot/manubot'! Will need to look further.

manubot/process/tests/test_ci.py
epogrebnyak: Actually looks better as two isolated tests with skipif, thanks for pointing.

manubot/process/tests/test_ci.py
epogrebnyak: linked to build systems with `@pytest.mark.skipif('TRAVIS' not in os.environ or 'APPVEYOR' not in os.environ, reason=...)`

manubot/process/tests/test_ci.py
```diff
@@ -9,50 +9,99 @@
...
+        """
+        assert info['repo_name'] == 'manubot'
+        assert info['repo_slug'].endswith('manubot')
+
+    @pytest.mark.skipif('TRAVIS' not in os.environ,
+                        reason="Test parameters are specific to Travis")    
+    def test_repo_slug_on_travis(self, info):
+        # TRAVIS_REPO_SLUG: The slug (in form: owner_name/repo_name) of the repository 
+        # currently being built. https://docs.travis-ci.com/user/environment-variables/
+        assert info['repo_slug'] == os.environ['TRAVIS_REPO_SLUG']
```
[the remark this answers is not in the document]
epogrebnyak: > ensure that repo_slug contains the actual owner_name/repo_name

Here I would like to stress the original goal of refactoring - removing hardcoded owner name for the tests, because it prevents from CI checks on a fork. I see three ways of doing this: 

1. pick up an owner name from an environment variable, make sure at least interface `info['repo_slug']` works
2. just assert `info['repo_slug']` is a string and ends with `/manubot` (standard repo name)
3. skip check for `info['repo_slug']` from the tests altogether

Which should we pick?

epogrebnyak:
> For example, I prefer the old tests, which are less the half the number of lines and have much less logic in them that can break and take readers extra time to understand.

Oversaturated logic is really bad for the tests I agree, we can close this PR and revert to original tests as a base + more forksfail tests .

The problem with the existing tests they are linked to a single repo, which prevents people from contributing to your project. One has to do PR's instead of testing in a fork branch to see what happens along with changes. If that is what maintainers want it should be stated somewhere in the docs or contributors manual "use PR's, the tests will fail on your fork".

I would argue it is a side of maintainers' responsibility to make a project structurally compatible with distributed work. Imagine `assert os.environ['TRAVIS_REPO_SLUG'] == 'manubot/manubot'` was not in your tests or it was sorted out by maintainers themselves in a way they prefer is the best. It would have left more contributors time to work on other issues, including user features. 

> we'll make the most progress if we apply your focus to areas of the codebase that will improve the user experience.

There is a limit in adding user features if your codebase is somewhat repetitive and unintuitive to follow. New features make code even more complicated, less transparency, less people who are willing to dig into the code - I think it is a vicious circle, not good for long-term viability of the project. I strongly believe #147 is a good attempt to address this challenge.

dhimmel:
> The problem with the existing tests they are linked to a single repo, which prevents people from contributing to your project.

Agree this is a problem and we should fix it. Since you're one of the first outside contributors to work with the codebase on such a deep level, you're the first one to hit many of these issues. We definitely want to fix them, such that subsequent contributors won't have to repeat your debugging and frustration!

> I would argue it is a side of maintainers' responsibility to make a project structurally compatible with distributed work.

Agree. Let me know ever if you'd like me to do the PR for any bugs you identify in issues. I'm happy to fix bugs like #140, but I generally like to give the person who discovers them the opportunity to fix them as a way of giving credit... especially when there is a straightforward solution. #140 turned out to be a bit more challenging than I envisioned, but hopefully the suggestion in https://github.com/manubot/manubot/pull/148#discussion_r333995364 will let us resolve it.

> New features make code even more complicated, less transparency, less people who are willing to dig into the code - I think it is a vicious circle, not good for long-term viability of the project. I strongly believe #147 is a good attempt to address this challenge.

Agree. Compared to many projects I've contributed to, I think this one is more willing to revisit existing implementations in hopes of maintainability and creating a scalable foundation. I'm with you that #147 is a promising architectural change that could do this.

epogrebnyak:
Superceded by #160.

PR #151, opened 2019-10-14
pytest options for caching data for local tests

epogrebnyak:
`pytest --use-requests-cache` will store API calls in `test_cache.sqlite` file and re-use that data within the next 12 hours when running tests. 

Usage:
```
pytest --use-requests-cache
```

Docstring, accessible with `pytest --fixtures`:

```
---------------------------------------------- fixtures defined from conftest -----------------------------------------------
use_requests_cache
    Use requests_cache for caching web API calls.
    Enabled with --use-requests-cache flag:

       pytest --use-requests-cache

    Notes:
    - The cache persists for 12 hours.
    - Cache does not working for invocations of manubot commands as subprocess.
```


conftest.py
```diff
@@ -0,0 +1,34 @@
+import pytest
```
dhimmel: Should this be located at `manubot/conftest.py` rather than in the root?

conftest.py
```diff
@@ -0,0 +1,34 @@
...
+    if request.config.getoption("--use-requests-cache"): 
+        start_caching(hours=12)
+
+
+def start_caching(**kwargs):
+    from datetime import timedelta
+    from requests_cache import install_cache
+    if not kwargs:
+        kwargs = dict(hours=1)
+    install_cache(cache_name='test_cache', # extension .sqlite always added by requests_cache
```
dhimmel: pytest seems to create a `.cache` directory where it caches some stuff. Perhaps we should place the cache at `.cache/requests-cache.sqlite`?

conftest.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    if request.config.getoption("--use-requests-cache"): 
+        start_caching(hours=12)
+
+
+def start_caching(**kwargs):
+    from datetime import timedelta
+    from requests_cache import install_cache
+    if not kwargs:
+        kwargs = dict(hours=1)
```
dhimmel: What if we just put `hours=1` or something similar in the function definition? This seems like extra logic that never ends up being used, given the rest of the implementation.

Unless is the idea that code elsewhere may call `start_caching`?

conftest.py
epogrebnyak: Indeed, `conftest.py` should be where `__init__.py` is.

conftest.py
epogrebnyak: pytest creates `.pytest_cache`, but its location is not exposed by pytest API. We can force-write
`.cache/requests-cache.sqlite` though.

conftest.py
dhimmel: https://github.com/manubot/manubot/pull/151/commits/a64ba13533181afddc53737f7c469935e44d0ec9

conftest.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--use-requests-cache",
+        action="store_true",
+        help="Use requests_cache for caching web API calls"
+    )
+    parser.addoption(
+        "--hours",
```
dhimmel: Let's rename to `--request-cache-hours`.

conftest.py
dhimmel: Still not sure why we can't just make `hours` an argument and remove the unused `**kwargs` logic and processing.

conftest.py
```diff
@@ -0,0 +1,56 @@
...
+def start_caching(**kwargs):
+    from datetime import timedelta
+    from requests_cache import install_cache
+    if not kwargs:
+        kwargs = dict(hours=1)
+    ttl = timedelta(**kwargs)
+    install_cache(cache_name=cache_path(), expire_after=ttl)
+    print('Using requests_cache to speed up local tests')
+    print('Cache location is', cache_path() + '.sqlite')
+    print('Cache expiration time is set to', ttl)
```
dhimmel: use `logging.info` rather than `print`

.gitignore
```diff
@@ -1,3 +1,6 @@
+# Database
+*.sqlite
+
```
dhimmel: This is no longer needed if the database goes in the `.cache` directory

conftest.py
```diff
@@ -0,0 +1,56 @@
...
+          pytest --use-requests-cache
+          pytest --use-requests-cache --hours=3
+          pytest --use-requests-cache --hours=8760
+
+       Notes:
+       - By default the cache persists for 12 hours. Default can be overrriden
+         with --hours option.
+       - Cache does not work for invocations of manubot commands as subprocess.
+
+       Cache file location: .\\.cache\\requests-cache.sqlite
```
dhimmel:
```suggestion
       Cache file location: .cache/requests-cache.sqlite
```

.gitignore
epogrebnyak: A path can possibly change, an extension is fixed, let's keep it.

epogrebnyak:
Note no tests use a fixture `request-cache-hours` yet in this branch, need to introduce the fixture in #146, after merged to master.

PR #154, opened 2019-10-14
Switch pytest to use verbose & colored output

dhimmel:


dhimmel:
@epogrebnyak any thoughts? I'm thinking the verbose output is worth it so we can see which tests ran on CI and which were skipped.

epogrebnyak:
For my workflow current test output is fine (on CI I'm interested if the tests pass or fail and the reason for failure - third party APIs or own code) but maybe verbose is useful too. We can try it and revert back if the output is too noisy.

PR #156, opened 2019-10-21
Migrate csl_item_passthrough to CSL_Item.clean

dhimmel:
Follows up on #150.

Migrates `csl_item_passthrough` standalone function to the `CSL_Item.clean` method.


dhimmel:
@epogrebnyak want to take a look? There is more to do in terms of adopting the CSL_Item class, but I want to do the migration as a series of smaller pull requests. This one is focused on the changes to remove the `csl_json_passthrough` function.

epogrebnyak:
Why not two separate methods - .set_id and .prune()? 

Pass_through was a combinator for different operations, should clean() be that way too? Maybe not - can use separate methods, and combine later.

epogrebnyak:
I think the idea of coupling two actions into `.clean()` comes from the behavior that schema check (pruning) will fail if ids are not set. I think this can be captured by logging or exceptions - the `.prune()` method can be verbose about the reason it fails, maybe even check the id's presence in specific.

epogrebnyak:
```python
    def prune(self):
        """
        Sanitize and touch-up a potentially dirty CSL_Item.
        The following steps are performed:
        - update incorrect values for "type" field when a correct variant is known
        - remove fields that violate the JSON Schema
        - set default value for "type" if missing, since CSL JSON requires type
        - validate against the CSL JSON schema to ensure output CSL_Item is clean
        """
        self.correct_invalid_type()
        self.prune_against_schema()
        self.set_default_type()
        return self

    def fix_types(self):
        # Similar to prune = False
        self.correct_invalid_type()
        self.set_default_type()
        return self
```

dhimmel:
> Why not two separate methods - `.set_id()` and `.prune()`? 

Yes on second thought, we can just call set_id prior to calling clean.

epogrebnyak:
In the call you can have: 

```
csl_item['id'] = shorten_citekey(standard_id) # even without a specific setter method
csl_item.prune()
```
instead of a former method

```
csl_item = csl_item_passthrough(csl_item, set_id=shorten_citekey(standard_id))
```

This arguably makes the transformations more granular, and unequivocal (we did a, then b).

epogrebnyak:
`prune_against_schema(self)` and `validate_against_schema(self)` are very short, clean, isolated.

epogrebnyak:
https://github.com/dhimmel/manubot/blob/70b6b3791abb3f9c4da41152bb71e89a72e0c08e/manubot/cite/csl_item.py#L110

`assert` will fail with some generic exception, maybe you might want more fanfare with custom exception or logging here.

dhimmel:
Removed set_id from .clean in https://github.com/manubot/manubot/pull/156/commits/889b0d34412d7495d232da7137abbeb1804c94c4

Regarding implementation proposed in https://github.com/manubot/manubot/pull/156#issuecomment-544691237. The readability is great, but it combines two processing steps that we have kept separate conceptually:

1. cleaning, which includes replacing dirty values and setting missing values. Cleaning can include pruning.
2. pruning, which is just removal of fields that violate the schema.

Therefore in https://github.com/manubot/manubot/pull/156/commits/889b0d34412d7495d232da7137abbeb1804c94c4 I kept .clean separate from .prune_against_schema. Something we can tweak in the future.

> assert will fail with some generic exception, maybe you might want more fanfare with custom exception or logging here.

This seemed like a pure usage of `assert` as per https://github.com/manubot/manubot/pull/146#discussion_r334003689:

> Asserts should be used to test conditions that should never happen. The purpose is to crash early in the case of a corrupt program state.

It should never happen that `csl_item is not self`, and I just want to quickly codify this expectation.

epogrebnyak:
> * cleaning, which includes replacing dirty values and setting missing values. Cleaning can include pruning.
> 
> * pruning, which is just removal of fields that violate the schema.

They are best separated by different fucntions or methods. Every time you set a bool switch you either call one function or another - maybe they need different names. 

I understand the switches in big interfaces like `pd.read_csv`, but in small functions you kind of obfuscate a choice between one behavior/transformation or another with a bool switch.  They are more explicit as separate/single responsibility functions. 

This is a bit mechanistic view, the mental model you worked for the transformation is equally important, so why not keep the clean() with a bool. The whole PR makes former csl_item_passthrough() a much better code, imo.

epogrebnyak:
@dhimmel himmel, do you want to tackle adding/reading notes from/to CSL_Item `append_to_csl_item_not()` next? I had a quick implementation in dataclasees branch, can move it to a separate PR. It still a clean-up, not much end use value though. 

I think there is more long-term value in Handle classes, like DOI, ISBN and similar - is that on your list too? Maybe some extra preparation / discussion needed before that change.

dhimmel:
> do you want to tackle adding/reading notes from/to CSL_Item append_to_csl_item_not() next?

I started working on that in #157. Will check out what you have in the dataclasses branch.

> I think there is more long-term value in Handle classes, like DOI, ISBN and similar - is that on your list too? Maybe some extra preparation / discussion needed before that change.

I'm interested in wrapping up the CSL_Item migration first, then assessing the Handle classes. But I think we should be able to wrap up the CSL_Item stuff pretty soon.

PR #160, opened 2019-10-25
skip test_get_continuous_integration_parameters_* on forks

dhimmel:
closes https://github.com/manubot/manubot/issues/140
supercedes https://github.com/manubot/manubot/pull/148

### DIFF
--- a/manubot/process/tests/test_ci.py
+++ b/manubot/process/tests/test_ci.py
@@ -10,6 +10,7 @@ from ..ci import (
 
 
 @pytest.mark.skipif('TRAVIS' not in os.environ, reason='tests environment variables set by Travis builds only')
+@pytest.mark.skipif(os.getenv('TRAVIS_REPO_SLUG') != 'manubot/manubot', reason='test fails on forks')
 def test_get_continuous_integration_parameters_travis():
     info = get_continuous_integration_parameters()
     assert info is not None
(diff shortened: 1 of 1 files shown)


dhimmel:
@epogrebnyak I think this is the simplest solution for #140. The downsides are that:

1. tests won't run on forks... not a big deal IMO
2. if the CI providers get rid of `TRAVIS_REPO_SLUG` or `APPVEYOR_REPO_NAME` test will be skipped and not fail. This is unlikely and we'd notice it.
3. If we rename the repo, we have to remember to change the skipif.

But I think these issues are pretty trivial and the simple solution is best.

epogrebnyak:
Indeed the simplest solution - one can separste tests that fail or not fail on a fork, but this is a good fix. 

The tests will be skipped locally because the environment vars are not set, I assume.

Maybe merge and I can try how it works on a fork?

dhimmel:
> The tests will be skipped locally because the environment vars are not set, I assume.

tests already skipped locally because 

```python
@pytest.mark.skipif('TRAVIS' not in os.environ, reason='tests environment variables set by Travis builds only')
```

But actually I could remove that because it's now redundant.

dhimmel:
> But actually I could remove that because it's now redundant.

Although it is perhaps somewhat more clear to have two skipif statements because it can give a more specific reason. Will merge as is.

Issue #152, opened 2019-10-14
Apply a consistent & standard style for Python formatting


epogrebnyak:
When submitting/reviewing PR's we spend a bit of time doing trivial changes of code formatting, which can be automated, eg with `autopep8` or `black`.

`autopep8` has an aggressive more which moves a lot of formatting: 

```
autopep8 -iaa filename
```

There is a softer version (just deals with trailing whitespace):

```
autopep8 -i filename --select W29
```

We need something in between to be able to apply regularly either just to new code or to all of codebase.

dhimmel:
I'm in favor of switching to black because it's the only one that makes almost all formatting decisions. From what I've seen, sometimes black will make bad decisions. In these cases, we can disable it with the `# fmt: on/off` comments.

We can blacken the codebase and enforce it via CI.

But let's try to merge all open PRs first that are not on hold.

epogrebnyak:
Answered as a question, can reopen later.

dhimmel:
Actually let's keep open until we implement a consistent code style to the repo. I think it's a good idea and perhaps we can do it after the current round of PRs (but perhaps before citekey classes)?

PR #164, opened 2019-10-28
Apply black formatting to Python codebase

dhimmel:
work in progress

closes https://github.com/manubot/manubot/issues/152


dhimmel:
@epogrebnyak what do you think about the diff? I question whether this really does anything valuable.

epogrebnyak:
Perhaps it shows your codebase is already well-formatted. Also some people are real serious about ' vs " and black fixes that. 

Black is needed for new submissions, to fix things like trailing newlines quickly. The fact it does little for existing codebase is good!

epogrebnyak:
Does `black --check .` apply to both travis and appv? What is planned next on this?

dhimmel:
> Does `black --check .` apply to both travis and appv?

Currently this PR only adds it to Travis. It runs on all travis jobs currently, although perhaps ideally it would only run on a single travis job, because we really only need to check once whether the formatting is correct.

> What is planned next on this?

I'm debating what to do about the cases where it makes the code less readable. Mostly this happens with `pytest.mark.parametrize` decorators where each it's most clear when each test is on its own line. Also it doesn't combine `([])`:

> image

The question is how much time to spend fighting it, versus just letting it do want it wants and accepting the slightly worse formatting.

dhimmel:
> The question is how much time to spend fighting it, versus just letting it do want it wants and accepting the slightly worse formatting.

I am started to realize that fighting black is futile. One must just accept what it demands and move on (:. @epogrebnyak I am okay with merging if you are. Will this mess you up on #147?

epogrebnyak:
I think the idea of black is that you cannot fight it, it is opinionated by default. With autopep8 there are more options (like exclude some type of formatting error), but tuning is dufficult too. 

Do you want to include some standard command to format a single file or all files to make or similar tool?

dhimmel:
```
Do you want to include some standard command to format a single file or all files to make or similar tool?
```

The command is straightforward to reformat everything: `black .` or a specific subset like `black conftest.py`. Perhaps I should add a note to the README to do this if black fails on CI.

epogrebnyak:
The comnands look pretty straightforward, but the notecwill be useful to
anyone not familiar with black. Some 'Developper guidelines' section in
README will be greatly appreciated.

In there we can add comments about the commands for testing locally
(excluding some travis-specific tests).

ср, 6 нояб. 2019 г., 18:04 Daniel Himmelstein <notifications@github.com>:

agitter:
Should we make a contributing file with this type of developer guidance?  Then [GitHub will link](https://help.github.com/en/github/building-a-strong-community/setting-guidelines-for-repository-contributors) to it from issues and we can link it from the main readme.

dhimmel:
I opened https://github.com/manubot/manubot/pull/171 to add more developer commands to the readme.

I'm not sure "contributor guidelines" is the same thing as "development commands" but I'm open to moving stuff here if we see a benefit to that.

Issue #165, opened 2019-11-01
UnboundLocalError: local variable 'process' referenced before assignment


olgabot:
Just created a repo this morning and followed the [SETUP](https://github.com/manubot/rootstock/blob/master/SETUP.md) instructions, including AppVeyorBot. But, the build is currently failing for master: https://travis-ci.com/czbiohub/primate-brain-organoid-paper/builds/134583403 (full deploy error message in details below) and the GitHub Pages HTML isn't loading: https://czbiohub.github.io/primate-brain-organoid-paper/

Is there a configuration step that I missed? There *was* an issue with setting the `MANUBOT_SSH_PRIVATE_KEY` (I didn't hit `Add` 🤦‍♀ ) but it is now set. Do you have any suggestions?

### Deploy error message

<details>

```
Deploying application
+set -o errexit -o nounset -o pipefail
++dirname czbiohub/primate-brain-organoid-paper
+OWNER_NAME=czbiohub
++basename czbiohub/primate-brain-organoid-paper
+REPO_NAME=primate-brain-organoid-paper
+export OWNER_NAME REPO_NAME
+envsubst
+mv webpage/README-complete.md webpage/README.md
+git config --global push.default simple
++git log --max-count=1 --format=%ae
+git config --global user.email daniel.himmelstein@gmail.com
++git log --max-count=1 --format=%an
+git config --global user.name 'Daniel Himmelstein'
+git checkout master
Previous HEAD position was a121e47... Update environment on 2019-10-31
Switched to branch 'master'
M	webpage/README.md
Your branch is up to date with 'origin/master'.
+git remote set-url origin git@github.com:czbiohub/primate-brain-organoid-paper.git
++ssh-agent -s
+eval 'SSH_AUTH_SOCK=/tmp/ssh-hEKmZD0aBpID/agent.8053; export SSH_AUTH_SOCK;
SSH_AGENT_PID=8054; export SSH_AGENT_PID;
echo Agent pid 8054;'
++SSH_AUTH_SOCK=/tmp/ssh-hEKmZD0aBpID/agent.8053
++export SSH_AUTH_SOCK
++SSH_AGENT_PID=8054
++export SSH_AGENT_PID
++echo Agent pid 8054
Agent pid 8054
+set +o xtrace
Identity added: (stdin) ((stdin))
+git remote set-branches --add origin gh-pages output
+git fetch origin gh-pages:gh-pages output:output
[...]

PR #166, opened 2019-11-01
Define `process` before referencing it

olgabot:
Maybe will fix the problem in https://github.com/manubot/manubot/issues/165 ?

### DIFF
--- a/manubot/webpage/webpage_command.py
+++ b/manubot/webpage/webpage_command.py
@@ -107,10 +107,10 @@ def checkout_existing_versions(args):
         '--',
         'v',
     ]
-    process = subprocess.run(command, stderr=subprocess.PIPE)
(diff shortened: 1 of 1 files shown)


manubot/webpage/webpage_command.py
```diff
@@ -107,10 +107,10 @@ def checkout_existing_versions(args):
         '--',
         'v',
     ]
+    process = subprocess.run(command, stderr=subprocess.PIPE)
     logging.info(
         f"Attempting checkout with the following command:\n{shlex_join(process.args)}"
```
dhimmel: Would be best if we keep the subprocess command after the logging statement but do:

```suggestion
        f"Attempting checkout with the following command:\n{shlex_join(command)}"
```

PR #169, opened 2019-11-05
Support metadata.thumbnail / variables.thumbnail_url

dhimmel:
Refs https://github.com/manubot/rootstock/issues/267

This commit adds a `thumbnail_url` field to `output/variables.json`. The thumbnail can be specified by a repo-relative path, an existing URL, or automatically detected if `thumbnail.png` exists.

git_repository_root is derived from _git_project_root in https://github.com/pdoc3/pdoc/commit/b47e916d78821dd9918b83f4a23b859b9cc473c5.

### DIFF
--- a/manubot/process/thumbnail.py
+++ b/manubot/process/thumbnail.py
@@ -1,0 +1,78 @@
+"""
+Tools for manuscript thumbnial detection and processing.
+"""
+import functools
+import pathlib
+import subprocess
+from urllib.parse import urlparse
+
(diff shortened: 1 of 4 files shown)


manubot/process/thumbnail.py
```diff
@@ -0,0 +1,81 @@
+"""
+Tools for manuscript thumbnial detection and processing.
```
agitter: `thumbnial` -> `thumbnail`

PR #171, opened 2019-11-07
readme: improve development commands

dhimmel:


dhimmel:
@epogrebnyak how does this look to you?

I thought about including some specific documentation of the pytest marks, but I actually think it may just be better to have the failure messages include those commands, since I don't like adding stuff to readme that easily gets outdated when we forget to update it with the code.

epogrebnyak:
The failure messages of specific tests, in docstring? Good idea, it will be more up to date than readme.

dhimmel:
> The failure messages of specific tests, in docstring? Good idea, it will be more up to date than readme.

Like you have at:

https://github.com/manubot/manubot/blob/0b7c14a6bbab77d6c91f0019dff524bb43587769/manubot/cite/tests/test_cite_command.py#L79-L82

Pytrest prints the docstrings for failed tests. I'm not sure whether the `Base_cite_command_render_stdout` docstring will display for failed tests, but the idea is that if so, it's the best place to suggest using `-m "not pandoc_version_sensitive"`.

README.md
```diff
@@ -226,32 +226,42 @@ optional arguments:
...
+### Commands
+
+Below are some common commands used for development.
+They assume the working directory is set to the repository's root, and the conda environment is activated. 
 
 ```shell
+# run the test suite
+pytest
+
+# reformat python files for according to the black style rules (required to pass CI)
```
dhimmel:
```suggestion
# reformat Python files according to the black style rules (required to pass CI)
```

Issue #172, opened 2019-11-08
process: reconsider metadata and variables distinction


dhimmel:
Currently we have a function `get_metadata_and_variables` that returns two dictionaries (`metadata` and `variables`) that is called during `manubot process`:

https://github.com/manubot/manubot/blob/9023b7fbfa0b235c14a4d702516bc0cd6d3101ed/manubot/process/util.py#L170-L220

`metadata` is written as a yaml block at the start of [`manuscript.md`](https://github.com/manubot/rootstock/blob/1fcd2834c080116fe544c7a7187b5b6d8dedfb81/manuscript.md). Presently rootstock has the following `metadata`:

```yaml
author-meta:
- John Doe
- Jane Roe
date-meta: '2019-11-01'
keywords:
- markdown
- publishing
- manubot
lang: en-US
title: Manuscript Title
```

`variables` currently contains more stuff and is written to [`variables.json`](https://github.com/manubot/rootstock/blob/1fcd2834c080116fe544c7a7187b5b6d8dedfb81/variables.json):

<details>
<summary>variables.json</summary>

```json
{
  "date": "November 1, 2019",
  "authors": [
    {
      "github": "johndoe",
      "name": "John Doe",
      "initials": "JD",
      "orcid": "XXXX-XXXX-XXXX-XXXX",
      "twitter": "johndoe",
      "email": "john.doe@something.com",
      "affiliations": [
        "Department of Something, University of Whatever"
      ],
      "funders": "Grant XXXXXXXX",
      "affiliation_numbers": [
        1
      ]
    },
    {
      "github": "janeroe",
      "name": "Jane Roe",
      "initials": "JR",
      "orcid": "XXXX-XXXX-XXXX-XXXX",
      "email": "jane.roe@whatever.edu",
      "affiliations": [
        "Department of Something, University of Whatever",
        "Department of Whatever, University of Something"
      ],
      "affiliation_numbers": [
        1,
        2
      ]
    }
  ],
  "affiliations": [
    {
      "affiliation": "Department of Something, University of Whatever",
      "affiliation_number": 1
    },
    {
      "affiliation": "Department of Whatever, University of Something",
      "affiliation_number": 2
    }
  ],
  "ci_source": {
    "provider": "travis",
    "repo_slug": "manubot/rootstock",
    "repo_owner": "manubot",
    "repo_name": "rootstock",
    "commit": "7cf9071212ce33116ad09cf2237a370b180a3c35",
[...]

PR #138, opened 2019-09-16
Set extra HTML <head> metadata fields using Pandoc's header-includes

dhimmel:
helpful references:

https://github.com/jgm/pandoc-templates/blob/master/default.html5
https://github.com/manubot/rootstock/issues/176
https://github.com/manubot/rootstock/issues/203
https://scholar.google.com/intl/en/scholar/inclusion.html#indexing
https://larsgw.blogspot.com/2018/07/journal-metadata-authors-institutions.html
https://ogp.me
https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/markup.html
https://www.monperrus.net/martin/accurate+bibliographic+metadata+and+google+scholar

PR #173, opened 2019-11-09
Combine pandoc metadata and jinja2 variables into a single dictionary

dhimmel:
closes https://github.com/manubot/manubot/issues/172

will require rootstock changes to frontmatter

### DIFF
--- a/manubot/process/util.py
+++ b/manubot/process/util.py
@@ -167,9 +167,17 @@ def add_author_affiliations(variables):
     return variables
 
 
-def load_metadata(args):
+def load_metadata(args) -> dict:
     """
     Process metadata.yaml and create variables available for jinja2 templating.
+
+    Returns a dictionary with the following keys:
+    - `pandoc`: a dictionary for passing options to Pandoc via the `yaml_metadata_block`.
+      All fields from a manuscript's `metadata.yaml` are copied to this dictionary,
+      expect for special Manubot-defined fields. Special fields include `author_info` and
+      `thumbnail`.
+    - `manubot`: a dictionary with manubot-generated variables.
+    - user-specified fields inserted according to the `--template-variables-path` option.
     """
     # Generated manuscript variables
     variables = {'pandoc': {}, 'manubot': {}}
--- a/manubot/process/util.py
+++ b/manubot/process/util.py
@@ -167,12 +167,12 @@ def add_author_affiliations(variables):
     return variables
 
 
-def get_metadata_and_variables(args):
+def load_metadata(args):
     """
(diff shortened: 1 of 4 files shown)


agitter:
Based on #172 and the partial implementation, this will only change the internal data structure and the `variables.json` file Manubot writes.  It won't change how users specify custom variables or metadata.  Is that correct?

dhimmel:
> It won't change how users specify custom variables or metadata. Is that correct?

That is correct. Specifying metadata will stay the same. However, code that consumes `variables.json` may have to be updated. For example, with the current implementation in https://github.com/manubot/manubot/issues/172, front-matter would have to be changed from using `authors` to `manubot.authors`.

dhimmel:
Note for when jinja2=2.11 is released: look into `jinja2.ChainableUndefined` such nested undefined variables in jinja2 templates won't raise `jinja2.exceptions.UndefinedError`. More info at https://github.com/pallets/jinja/pull/997 & https://github.com/pallets/jinja/issues/811.

dhimmel:
Okay I am going to merge this since I think it's a crucial precursor to the more advanced metadata insertion we're implementing in #138. Manuscript frontmatters will have to be updated once rootstock updates to this version. CI will generally fail with `jinja2.exceptions.UndefinedError` unless front-matter is updated, which has the nice benefit that users are unlikely to start having corrupted front-matters without noticing.

manubot/process/util.py
```diff
@@ -168,12 +168,21 @@ def add_author_affiliations(variables):
...
+    - `pandoc`: a dictionary for passing options to Pandoc via the `yaml_metadata_block`.
+      All fields from a manuscript's `metadata.yaml` are copied to this dictionary,
+      expect for special Manubot-defined fields. Special fields include `author_info` and
+      `thumbnail`.
+    - `manubot`: a dictionary with manubot-generated variables.
+    - user-specified fields inserted according to the `--template-variables-path` option.
     """
     # Generated manuscript variables
-    variables = collections.OrderedDict()
+    variables = {"pandoc": {}, "manubot": {}}
```
agitter: @dhimmel are these not OrderedDict because the regular dict is ordered in Python 3.6+?

dhimmel: Yes and now Python has committed to keeping dicts ordered in future versions.

Before, our practice was to use OrderedDict to ensure dict order. But now since that's gauranteed going forward, we should use OrderedDict only when necessary to trigger a special behavior somewhere (like pandas sometimes treats dicts and ordereddicts separately). But I'm thinking we should default to using dict and only cast to OrderedDict when absolutely necessary. I imagine OrderedDict may become deprecated someday.

PR #181, opened 2019-12-02
process: add software versions to variables

dhimmel:
refs https://github.com/manubot/rootstock/issues/282

### DIFF
--- a/manubot/process/metadata.py
+++ b/manubot/process/metadata.py
@@ -159,7 +159,16 @@ def get_manuscript_urls(html_url: Optional[str] = None) -> dict:
     return urls
 
 
-def get_software_versions():
+def get_software_versions() -> dict:
+    """
+    Return a dictionary of software versions for softwares components:
+
+    - manubot_version: the semantic version number of the manubot python package.
+    - rootstock_commit: the version of the rootstock repository, as a commit hash,
+      included in the manuscript repository.
+
+    Values whose detection fails are set to None.
+    """
     from manubot import __version__ as manubot_version
 
     return {
@@ -168,7 +177,7 @@ def get_software_versions():
     }
 
 
-def get_rootstock_commit():
+def get_rootstock_commit() -> Optional[str]:
     """
     Return the most recent commit in common between the git repository
     this function is run within (usually a Manubot manuscript repository)
--- a/manubot/process/metadata.py
+++ b/manubot/process/metadata.py
@@ -200,7 +200,7 @@ def get_rootstock_commit():
         )
     except subprocess.CalledProcessError as error:
         logging.warning(
-            f"get_rootstock_version: {shlex_join(error.cmd)!r} returned exit code {error.returncode} "
+            f"get_rootstock_commit: {shlex_join(error.cmd)!r} returned exit code {error.returncode} "
             f"with the following stdout:\n{error.stdout}\n"
             f"And the following stderr:\n{error.stderr}"
         )
--- a/manubot/process/metadata.py
+++ b/manubot/process/metadata.py
@@ -157,3 +157,53 @@ def get_manuscript_urls(html_url: Optional[str] = None) -> dict:
             + "\n".join(x.url for x in response.history + [response])
         )
     return urls
+
+
+def get_software_versions():
+    from manubot import __version__ as manubot_version
+
+    return {
+        "manubot_version": manubot_version,
+        "rootstock_commit": get_rootstock_commit(),
+    }
+
+
+def get_rootstock_commit():
+    """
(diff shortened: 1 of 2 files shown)


dhimmel:
Currently, `manubot_version` is detected from:

https://github.com/manubot/manubot/blob/ddf418b3510a719e2973a066dcc2e066c13c373f/manubot/__init__.py#L1

But since we install `manubot` package by the commit, this is only a rough version number. We could in addition add a `manubot_commit` variable which detects the commit. Numpy has something like this in their [`setup.py`](https://github.com/numpy/numpy/blob/f434a6444304aff2c937da93e2f2045ba4419f05/setup.py#L129-L138).

dhimmel:
Failing due to some arxiv API issue. https://export.arxiv.org/api/query?id_list=1806.05726v1&max_results=1 works for me in my browser but on CI is [encountering](https://travis-ci.com/manubot/manubot/jobs/262309967#L871):

```
HTTPSConnectionPool(host='export.arxiv.org', port=443): Max retries exceeded with url: /api/query?id_list=1806.05726v1&max_results=1 (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1108)')))
```

I tested this locally that running this version of manubot on a manuscript repo works as expected to set `manubot_version` and `rootstock_commit`. A bit difficult to thoroughly test in this repo unless we were to make submodules or other complexities.

manubot/process/metadata.py
```diff
@@ -157,3 +157,53 @@ def get_manuscript_urls(html_url: Optional[str] = None) -> dict:
...
+    try:
+        args = ["git", "fetch", "rootstock", "master"]
+        subprocess.check_output(args, stderr=subprocess.PIPE, universal_newlines=True)
+        args = ["git", "merge-base", "master", "rootstock/master"]
+        output = subprocess.check_output(
+            args, stderr=subprocess.PIPE, universal_newlines=True
+        )
+    except subprocess.CalledProcessError as error:
+        logging.warning(
+            f"get_rootstock_version: {shlex_join(error.cmd)!r} returned exit code {error.returncode} "
```
agitter: Do you want `get_rootstock_commit` here?

manubot/process/metadata.py
```diff
@@ -157,3 +157,53 @@ def get_manuscript_urls(html_url: Optional[str] = None) -> dict:
...
+    }
+
+
+def get_rootstock_commit():
+    """
+    Return the most recent commit in common between the git repository
+    this function is run within (usually a Manubot manuscript repository)
+    and the `master` branch of the `rootstock` remote.
+
+    WARNING: This function may modify the git repository its executed within:
```
agitter: How conservative do we want to be regarding modifying the local git repository?  This approach is reasonable.  An even more conservative approach would be to only run this if 1) the `rootstock` remote already exists or 2) Manubot is being run in a CI service.

manubot/process/metadata.py
```diff
@@ -157,3 +157,53 @@ def get_manuscript_urls(html_url: Optional[str] = None) -> dict:
             + "\n".join(x.url for x in response.history + [response])
         )
     return urls
+
+
+def get_software_versions():
```
agitter: Once this is stable, can you please add some light documentation for the auto-generated docs?

manubot/process/metadata.py
```diff
@@ -157,3 +157,53 @@ def get_manuscript_urls(html_url: Optional[str] = None) -> dict:
...
+    from manubot import __version__ as manubot_version
+
+    return {
+        "manubot_version": manubot_version,
```
agitter: This seems like a good starting point.  If you want to expand the options in the future, you could still keep `manubot_version` as is and add another key like `manubot_commit`.

manubot/process/metadata.py
dhimmel: > How conservative do we want to be regarding modifying the local git repository? 

I don't think it's too damaging to create the `rootstock` remote if it doesn't exist. I think it's important that we keep the functionality for local builds not just CI. I will think if there's a better way we can ensure `rootstock` remote is only added if the repo seems to be a manubot manuscript.

agitter: Sure, that makes sense.  I wanted to open the discussion but don't have any objections.

This being part of the `process` subcommand may already be sufficient for making sure it is only run as part of a Manubot manuscript.

dhimmel: I think we can proceed with the current implementation and continue to think if there is a more targeted solution.

Issue #61, opened 2018-09-23
URL to citation


agitter:
In https://github.com/kipoi/website/issues/77 @dhimmel and @Avsecz discussed a potential new Manubot feature.  Per @dhimmel:
> One potential Manubot feature that could be helpful would be "URL to citation" that could perform the following conversion:

> https://doi.org/10.1101/gr.227819.117 ⟶ doi:10.1101/gr.227819.117
http://dx.doi.org/10.1101/gr.227819.117 ⟶ doi:10.1101/gr.227819.117
https://doi.org/gd54x8 ⟶ doi:10.1101/gr.227819.117
https://arxiv.org/pdf/1603.09123.pdf ⟶ arxiv:1603.09123

> Would also hopefully work with other types of URLs.

dhimmel:
There are two possible features that may be useful:

1. reformatting URLs to be citation stings, as shown above. In these cases, the URLs always contain the citation identifier, but just need the identifier to be extracted and reformatted.
2. returning a list of citation strings that would cite the work referred to by the URL.

For the second, passing the URL `https://peerj.com/articles/705/` would return the following list:

0. `doi:10.7717/peerj.705`
1. `pmcid:PMC4304851`
2. `pmid:25648772`
3. `url:https://peerj.com/articles/705/`

I ordered the list by manubot's citation-source preference.

The http://kipoi.org/ use case of using Manubot to get authors for a URL reference would probably be OKAY with either method.

Chime in if you think either method would be useful... we probably won't pursue this unless there is some user interest.

Avsecz:
For Kipoi, the both features are useful. Second one could be used for displaying the number of papers citing the model on the website (and maybe even link to a page where these papers are listed).

For the start, I think supporting URLs from doi, pubmed, arxiv and bioarxiv would be enough as these cover a broad spectrum. One can start with the functions I wrote for the website: https://github.com/kipoi/website/blob/master/app/models/authors.py#L6

PR #182, opened 2019-12-11
url_to_citekey: convert from URL to citekey

dhimmel:
closes https://github.com/manubot/manubot/issues/61

currently, `url_to_citekey` is not used by Manubot, but is a convenience function for users of the package.

todo:

- [x] should we standardize?
- [x] should we handle errors?
- [x] should we test pathological URLs?

### DIFF
--- a/manubot/cite/citekey.py
+++ b/manubot/cite/citekey.py
@@ -323,6 +323,10 @@ def url_to_citekey(url):
         # DOI URLs
         doi = unquote(parsed_url.path.lstrip("/"))
         return f"doi:{doi}"
+    if parsed_url.hostname.split(".")[-2:] == ["biorxiv", "org"]:
+        # bioRxiv URL to DOI
+        (biorxiv_id,) = re.findall(r"/([0-9]{6})", parsed_url.path)
+        return f"doi:10.1101/{biorxiv_id}"
     is_ncbi_url = parsed_url.hostname.endswith("ncbi.nlm.nih.gov")
     if is_ncbi_url and parsed_url.path.startswith("/pubmed/"):
         # PubMed URLs
--- a/manubot/cite/citekey.py
+++ b/manubot/cite/citekey.py
@@ -86,6 +86,9 @@ def standardize_citekey(citekey, warn_if_changed=False):
 
 
 regexes = {
+    "arxiv": re.compile(
+        r"([0-9]{4}\.[0-9]{4,5}|[a-z\-]+(\.[A-Z]{2})?/[0-9]{7})(v[0-9]+)?"
+    ),
     "pmid": re.compile(r"[1-9][0-9]{0,7}"),
     "pmcid": re.compile(r"PMC[0-9]+"),
     "doi": re.compile(r"10\.[0-9]{4,9}/\S+"),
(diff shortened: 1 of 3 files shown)


manubot/cite/citekey.py
```diff
@@ -309,3 +309,44 @@ def infer_citekey_prefix(citekey):
...
+
+    parsed_url = urlparse(url)
+    if parsed_url.hostname.split(".")[-2:] == ["doi", "org"]:
+        # DOI URLs
+        doi = unquote(parsed_url.path.lstrip("/"))
+        return f"doi:{doi}"
+    if parsed_url.hostname.split(".")[-2:] == ["biorxiv", "org"]:
+        # bioRxiv URL to DOI
+        (biorxiv_id,) = re.findall(r"/([0-9]{6})", parsed_url.path)
+        return f"doi:10.1101/{biorxiv_id}"
```
dhimmel: @rabdill this is snippet to extract the DOI for a preprint from its biorxiv URL. I couldn't find much documentation on bioRxiv IDs and how the URL corresponds to the DOI. Do you know if the regex is correct? Are all IDs 6 digits? Would matching on 6+ digits be more future proof (will bioRxiv need to start issuing 7 digit IDs in future)?

rabdill: I haven't seen any documentation either. I'm assuming you're referring to the URI structure that looks like `"/content/early/2019/01/22/${id}"`, right? As far as I can tell, all the canonical URLs now use the new format, `/content/10.1101/${id}v1"`, where the DOI is basically the last thing in the URL before the version identifier. Does that simplify anything for you, or do you want to support both URL formats?

In any case, this regex matches all the DOIs that are currently indexed, but I'd guess it could jump to 7 digits at some point—I think the `10.1101` prefix is used by other stuff from Cold Spring Harbor Labs as well. There's also no reason I know of that they'd stay numeric, either. Would taking everything before the version indicator be too liberal? i.e. `r"(10.1101/\w+)v\d+$"`

dhimmel: Thanks @rabdill for the insights, super helpful.

> Does that simplify anything for you, or do you want to support both URL formats?

Would like to support the following URLs from the tests (all of which were generated as of yesterday):

https://github.com/manubot/manubot/blob/d9f4676f485e1e0f09047437f8cf6d6d25cbe345/manubot/cite/tests/test_citekey.py#L134-L139

Note the PDF link still uses the archaic URL format. 

> There's also no reason I know of that they'd stay numeric, either. Would taking everything before the version indicator be too liberal?

I'm worried about the URLs where `10.1101/` does not prefix the identifier. Maybe for now I'll change the regex to match 6+ digits to allow for expansion of the identifier space. Can always update later if alphabetical characters are added to identifiers. The nature of this function is that it's okay to fail to extract the DOI, but its not okay to extract an incorrect DOI.

rabdill: Well, that didn't take long. In case you didn't catch this, looks like the DOI pattern just changed: https://twitter.com/cshperspectives/status/1205109455752454144

dhimmel: So new URLs are like:

```
https://www.biorxiv.org/content/10.1101/2019.12.11.872580v1
https://www.biorxiv.org/content/10.1101/2019.12.11.872580v1.full.pdf+html
https://www.biorxiv.org/content/10.1101/2019.12.11.872580v1.full.pdf
```

Old URLs have not changed. The implementation would be simpler if we dropped support for legacy PDF links like `https://www.biorxiv.org/content/early/2017/08/31/087619.full.pdf`.

dhimmel:
> should we standardize?

Will leave standardization up to the users. 

> should we handle errors? should we test pathological URLs?

This is now mostly handled. Weird inputs usually are kept as URLs rather than converted to broken identifiers.

PR #188, opened 2019-12-18
process: YAML metadata use 'authors' not 'author_info'

dhimmel:
'authors' as a key is more consistent with other YAML metadata
frontmatters in the space.

### DIFF
--- a/manubot/process/tests/manuscripts/example/content/metadata.yaml
+++ b/manubot/process/tests/manuscripts/example/content/metadata.yaml
@@ -2,7 +2,7 @@
 title: "Example manuscript for testing"
 keywords:
   - manubot
-author_info:
+authors:
   -
     github: dhimmel
     name: Daniel S. Himmelstein
--- a/manubot/process/util.py
+++ b/manubot/process/util.py
@@ -175,7 +175,7 @@ def load_variables(args) -> dict:
       includes a `manubot` dictionary.
     - All fields from a manuscript's `metadata.yaml` that are not interpreted by Manubot are
       copied to `variables`. Interpreted fields include `pandoc`, `manubot`, `title`,
-      `keywords`, `author_info`, `lang`, and `thumbnail`.
+      `keywords`, `authors` (formerly `author_info`, now deprecated), `lang`, and `thumbnail`.
     - User-specified fields inserted according to the `--template-variables-path` option.
       User-specified variables take highest precedence and can overwrite values for existing
       keys like `pandoc` or `manubot` (dangerous).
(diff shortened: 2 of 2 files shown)


adebali:
I guess, after this change, I cannot output the author information with the latest rootstock which has `authors` instead of `author_info`. I tried to install the latest version of manubot but it didn't help. Any piece of advice on upgrading to the latest version of the tool?

dhimmel:
> Any piece of advice on upgrading to the latest version of the tool?

A little more information on where you are running manubot will help: one CI or locally?

Sounds like you've already updated your manuscript with the latest rootstock commits as per [these instructions](https://github.com/manubot/rootstock/blob/master/SETUP.md#merging-upstream-rootstock-changes)?

If running locally, you will likely have to update your conda environment with one of the commands [here](https://github.com/manubot/rootstock/tree/master/build#environment).

adebali:
I run manubot locally. 

Everything worked smoothly with the previous manuscripts. I started a new one by cloning the fresh rootstock. However, it doesn't go into the following loop:

```
{% for author in authors %}
{% if author.orcid is defined -%}
[![ORCID icon](images/orcid.svg){height="11px" width="11px"}](https://orcid.org/{{author.orcid}})
{%- endif %}
{{author.name}}<sup>{{author.affiliation_numbers | join(',')}}{%- if author.symbol_str is defined -%},{{author.symbol_str}}{%- endif -%}</sup>
{%- if not loop.last -%}, {%- endif -%}
{% endfor %}
```

I tried this:

```
conda env remove --name manubot
conda env create --file build/environment.yml
conda activate manubot
```

Still, no author information.

dhimmel:
> Everything worked smoothly with the previous manuscripts. I started a new one by cloning the fresh rootstock.

Ah but you kept the front-matter from the previous manuscript? I think the cause of your problem is that you're missing the front-matter changes in https://github.com/manubot/rootstock/commit/f1bf210b115607b1887c0dcfaed11cbe81543967. Specifically, the authors line becomes:

```
{% for author in manubot.authors %}
```

If you haven't done any manual edits to the front-matter, I'd just take the entire front-matter from the rootstock version you used to create the new manuscript.

I opened a rootstock PR at https://github.com/manubot/rootstock/pull/298 to make the upgrade instructions clearer. The upgrade process there will help avoid these issues.

In you `metadata.yaml` for the new manuscript, can you double check that `authors` is used instead of `author_info` (although actually I think we currently support both)?

Also are you getting any warnings in the terminal output or CI logs with the new manuscript?

Finally, what is in your output/variables

adebali:
Yes! It solved it. I copied and pasted the front-matter context from a previous version. That was the reason.

Yes, `authors` is used instead of `authors_info`. 

I didn't get any warnings.

> Finally, what is in your output/variables

Not sure what you mean by the variables.

dhimmel:
> Yes! It solved it. I copied and pasted the front-matter context from a previous version. That was the reason.

Awesome! 

> Not sure what you mean by the variables.

That was a fragment of an incomplete sentence before I realized what the actual issue was... disregard.

PR #189, opened 2019-12-30
CSL_Item: add date IO functionality

dhimmel:
~~work in progress. need to test and decide whether this is too complex / whether we should rely on a third party package like dateutil~~

### DIFF
--- a/manubot/cite/csl_item.py
+++ b/manubot/cite/csl_item.py
@@ -371,7 +371,7 @@ def date_parts_to_string(date_parts, fill: bool = False) -> Optional[str]:
         if not isinstance(part, str):
             break
         part = part.zfill(width)
-        if not len(part) != width or not part.isdigit():
+        if len(part) != width or not part.isdigit():
             break
         str_parts.append(part)
     if not str_parts:
--- a/manubot/cite/csl_item.py
+++ b/manubot/cite/csl_item.py
@@ -367,8 +367,11 @@ def date_parts_to_string(date_parts, fill: bool = False) -> Optional[str]:
     for i, part in enumerate(date_parts[:3]):
         width = widths[i]
         if isinstance(part, int):
-            part = str(part).zfill(width)
-        if not isinstance(part, str) or len(part) != width or not part.isdigit():
+            part = str(part)
+        if not isinstance(part, str):
+            break
+        part = part.zfill(width)
+        if not len(part) != width or not part.isdigit():
             break
         str_parts.append(part)
     if not str_parts:
--- a/manubot/cite/csl_item.py
+++ b/manubot/cite/csl_item.py
@@ -26,6 +26,7 @@ for csl_data that is JSON-serialized.
 import copy
 import logging
 import re
+from typing import List, Optional
 
 from manubot.cite.citekey import (
     standardize_citekey,
@@ -152,18 +153,18 @@ class CSL_Item(dict):
             self.validate_against_schema()
         return self
 
-    def set_date_variable(self, iso_date, variable="issued"):
+    def set_date(self, date, variable="issued"):
         """
-        iso_date: date either as a string (in the form YYYY, YYYY-MM, or YYYY-MM-DD)
+        date: date either as a string (in the form YYYY, YYYY-MM, or YYYY-MM-DD)
             or as a Python date object (datetime.date or datetime.datetime).
-        variable: which variable to assing the date to.
+        variable: which variable to assign the date to.
         """
-        date_parts = date_to_date_parts(iso_date)
+        date_parts = date_to_date_parts(date)
         if date_parts:
             self[variable] = {"date-parts": [date_parts]}
         return self
 
-    def get_date_variable(self, variable="issued", fill=False):
+    def get_date(self, variable="issued", fill=False):
         """
         Return a CSL date-variable as ISO formatted string:
         ('YYYY', 'YYYY-MM', 'YYYY-MM-DD', or None).
@@ -303,22 +304,22 @@ def assert_csl_item_type(x):
         raise TypeError(f"Expected CSL_Item object, got {type(x)}")
 
 
-def date_to_date_parts(iso_date):
+def date_to_date_parts(date) -> Optional[List[int]]:
     """
     Convert a date string or object to a date parts list.
 
-    iso_date: date either as a string (in the form YYYY, YYYY-MM, or YYYY-MM-DD)
+    date: date either as a string (in the form YYYY, YYYY-MM, or YYYY-MM-DD)
         or as a Python date object (datetime.date or datetime.datetime).
     """
     import datetime
 
-    if iso_date is None:
+    if date is None:
         return None
-    if isinstance(iso_date, (datetime.date, datetime.datetime)):
-        iso_date = iso_date.isoformat()
-    if not isinstance(iso_date, str):
-        raise ValueError(f"date_to_date_parts: unsupported type for {iso_date}")
-    iso_date = iso_date.strip()
+    if isinstance(date, (datetime.date, datetime.datetime)):
+        date = date.isoformat()
+    if not isinstance(date, str):
+        raise ValueError(f"date_to_date_parts: unsupported type for {date}")
+    date = date.strip()
     re_year = r"(?P<year>[0-9]{4})"
     re_month = r"(?P<month>1[0-2]|0[1-9])"
     re_day = r"(?P<day>[0-3][0-9])"
@@ -329,7 +330,7 @@ def date_to_date_parts(iso_date):
         f".*",  # regex to match anything
     ]
     for pattern in patterns:
-        match = re.match(pattern, iso_date)
+        match = re.match(pattern, date)
         if match:
             break
     date_parts = []
@@ -345,12 +346,13 @@ def date_to_date_parts(iso_date):
         return date_parts
 
 
-def date_parts_to_string(date_parts, fill: bool = False):
+def date_parts_to_string(date_parts, fill: bool = False) -> Optional[str]:
     """
     Return a CSL date-parts list as ISO formatted string:
     ('YYYY', 'YYYY-MM', 'YYYY-MM-DD', or None).
 
     date_parts: list or tuple like [year, month, day] as integers.
+        Also supports [year, month] and [year] for situations where the day or month-and-day are missing.
     fill: if True, set missing months to January
         and missing days to the first day of the month.
     """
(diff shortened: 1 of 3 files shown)


agitter:
How much of this functionality could dateutil provide?  Relying on a dedicated package instead of implementing and testing all of the possibilities within manubot is somewhat appealing.

dhimmel:
> How much of this functionality could dateutil provide?

[`dateutil.parser.isoparser`](https://dateutil.readthedocs.io/en/stable/_modules/dateutil/parser/isoparser.html) can create datetime objects for `YYYY`, `YYYY-MM`, and `YYYY-MM-DD` as well as other weird forms of ISO-8601 dates. But it always fills in the month and day even if just given `YYYY`.

> Relying on a dedicated package instead of implementing and testing all of the possibilities within manubot is somewhat appealing.

Agree. But now that I added tests, I'm a bit less concerned about the complexity of the implementations.

I didn't find much existing infrastructure that handles the case we have where we usually do not want to fill in missing months/days.

agitter:
> I didn't find much existing infrastructure that handles the case we have where we usually do not want to fill in missing months/days.

That's a good reason for a custom implementation.  Is this still a work in progress or ready for review?

dhimmel:
> Is this still a work in progress or ready for review?

ready for review, especially whether the function names are best. We could do `CSL_Item.get_date()` and `CSL_Item.set_date()` as simpler options to `*_date_variable`.

manubot/cite/csl_item.py
```diff
@@ -152,6 +152,32 @@ def clean(self, prune: bool = True):
             self.validate_against_schema()
         return self
 
+    def set_date_variable(self, iso_date, variable="issued"):
+        """
+        iso_date: date either as a string (in the form YYYY, YYYY-MM, or YYYY-MM-DD)
+            or as a Python date object (datetime.date or datetime.datetime).
+        variable: which variable to assing the date to.
```
agitter: assing -> assign

manubot/cite/csl_item.py
```diff
@@ -152,6 +152,32 @@ def clean(self, prune: bool = True):
             self.validate_against_schema()
         return self
 
+    def set_date_variable(self, iso_date, variable="issued"):
```
agitter: Should we look to the `note` as another option for how this setter should work?
```python
@note.setter
    def note(self, text: str):
```

manubot/cite/csl_item.py
agitter: assing -> assign

manubot/cite/csl_item.py
```diff
@@ -152,6 +152,32 @@ def clean(self, prune: bool = True):
...
+        iso_date: date either as a string (in the form YYYY, YYYY-MM, or YYYY-MM-DD)
+            or as a Python date object (datetime.date or datetime.datetime).
+        variable: which variable to assing the date to.
+        """
+        date_parts = date_to_date_parts(iso_date)
+        if date_parts:
+            self[variable] = {"date-parts": [date_parts]}
+        return self
+
+    def get_date_variable(self, variable="issued", fill=False):
```
agitter: This behavior is probably too complex to consider the date as a property, right?

manubot/cite/csl_item.py
```diff
@@ -275,3 +301,75 @@ def standardize_id(self):
 def assert_csl_item_type(x):
     if not isinstance(x, CSL_Item):
         raise TypeError(f"Expected CSL_Item object, got {type(x)}")
+
+
+def date_to_date_parts(iso_date):
```
agitter: Would we ever need this as a util function outside of this class?

manubot/cite/csl_item.py
```diff
@@ -275,3 +301,75 @@ def standardize_id(self):
...
+    if iso_date is None:
+        return None
+    if isinstance(iso_date, (datetime.date, datetime.datetime)):
+        iso_date = iso_date.isoformat()
+    if not isinstance(iso_date, str):
+        raise ValueError(f"date_to_date_parts: unsupported type for {iso_date}")
+    iso_date = iso_date.strip()
+    re_year = r"(?P<year>[0-9]{4})"
+    re_month = r"(?P<month>1[0-2]|0[1-9])"
+    re_day = r"(?P<day>[0-3][0-9])"
```
agitter: This could be a slightly narrower range, similar to how you excluded some double digit numbers from the month regex.  But do we actually need to check for valid dates here?

manubot/cite/csl_item.py
```diff
@@ -275,3 +301,75 @@ def standardize_id(self):
 def assert_csl_item_type(x):
     if not isinstance(x, CSL_Item):
         raise TypeError(f"Expected CSL_Item object, got {type(x)}")
+
+
+def date_to_date_parts(iso_date):
+    """
+    Convert a date string or object to a date parts list.
```
agitter: It would be helpful to define a date parts list as a list of integers.

manubot/cite/csl_item.py
```diff
@@ -275,3 +301,75 @@ def standardize_id(self):
...
+    if date_parts:
+        return date_parts
+
+
+def date_parts_to_string(date_parts, fill: bool = False):
+    """
+    Return a CSL date-parts list as ISO formatted string:
+    ('YYYY', 'YYYY-MM', 'YYYY-MM-DD', or None).
+
+    date_parts: list or tuple like [year, month, day] as integers.
```
agitter: Should we mention the month and day are optional?

manubot/cite/csl_item.py
```diff
@@ -275,3 +301,75 @@ def standardize_id(self):
...
+    if not isinstance(date_parts, (tuple, list)):
+        raise ValueError(f"date_parts must be a tuple or list")
+    while fill and 1 <= len(date_parts) < 3:
+        date_parts.append(1)
+    widths = 4, 2, 2
+    str_parts = []
+    for i, part in enumerate(date_parts[:3]):
+        width = widths[i]
+        if isinstance(part, int):
+            part = str(part).zfill(width)
```
agitter: Could this cause unexpected behavior if somehow a year is stored as `YY` (e.g. `98`) and is then zero filled?

manubot/cite/tests/test_csl_item.py
```diff
@@ -179,3 +185,46 @@ def test_csl_item_note_append(input_note, text, dictionary, expected_note):
...
+@pytest.mark.parametrize(
+    ["date", "expected"],
+    [
+        (None, None),
+        ("", None),
+        ("2019", [2019]),
+        ("2019-01", [2019, 1]),
+        ("2019-12", [2019, 12]),
+        ("2019-12-31", [2019, 12, 31]),
+        ("2019-12-99", [2019, 12]),
```
agitter: Related to the above comment, we could also watch for `2019-12-39` as an invalid day.

manubot/cite/tests/test_csl_item.py
```diff
@@ -179,3 +185,46 @@ def test_csl_item_note_append(input_note, text, dictionary, expected_note):
...
+        (None, [], False),
+        ("2019", [2019], False),
+        ("2019-01-01", [2019], True),
+        ("2019-01", [2019, 1], False),
+        ("2019-12", [2019, 12], False),
+        ("2019-12-01", [2019, 12], True),
+        ("2019-12-31", [2019, 12, 31], False),
+        ("2019-12", [2019, 12, "bad day"], False),
+        ("2019-12-01", [2019, 12, 1], False),
+        ("2019-12-01", ["2019", "12", "01"], False),
```
agitter: `["2019", "12", "1"]` could be another good test case

manubot/cite/tests/test_csl_item.py
```diff
@@ -179,3 +185,46 @@ def test_csl_item_note_append(input_note, text, dictionary, expected_note):
...
+    ["expected", "date_parts", "fill"],
+    [
+        (None, None, False),
+        (None, [], False),
+        ("2019", [2019], False),
+        ("2019-01-01", [2019], True),
+        ("2019-01", [2019, 1], False),
+        ("2019-12", [2019, 12], False),
+        ("2019-12-01", [2019, 12], True),
+        ("2019-12-31", [2019, 12, 31], False),
```
agitter: Change this to `True`?  It looks like it is testing the same condition as
```python
("2019-12-30", [2019, 12, 30], False)
```

manubot/cite/csl_item.py
dhimmel: These setters only work for set functions that only take a single positional argument. Therefore, a single `@date.setter` would be unable to set `variable="issued"`. We could eventually add properties with setters like `date_issued`, but not a huge benefit for the extra code

manubot/cite/csl_item.py
dhimmel: With the `variable` and `fill` options, I think yes... cannot fit within a property.

manubot/cite/csl_item.py
dhimmel: Yes I think it's useful to have access to it without having to create a dummy CSL_Item, so I placed it outside the class.

manubot/cite/csl_item.py
dhimmel: I am okay with `[99, 1, 1]` becoming `0099-01-01`. Rarely a reference may be to a work of that age. I'd say it's not our problem... users should pass full years. For example, would `10` be 1910 or 2010... I don't think we should make inferences here.

manubot/cite/csl_item.py
dhimmel: Yeah the function does a limited and incomplete check for date validity. I am okay with this for now. We haven't really decided where date validity checking should go, if anywhere.

manubot/cite/tests/test_csl_item.py
dhimmel: I am okay with not detecting `2019-12-39` as invalid. We don't gaurantee that we catch invalid dates, although we try to avoid extracting them, we're not actually verifying.

dhimmel:
@agitter I think I addressed your comments in https://github.com/manubot/manubot/pull/189/commits/31f599a49c1d128c031efdb5609ed10bf38931a5

manubot/cite/csl_item.py
agitter: I'm okay with this behavior too.  It could be a good test case as a way of documenting that is the expected behavior.

agitter:
The revisions look good to me once the tests pass.  I added one other suggestion above for another test case.

PR #190, opened 2020-01-04
process.util: improve modularity of get_citekeys_df and generate_csl_items

dhimmel:
I've resumed work on #99, which needs to access some functions from manubot.cite.util in a more general purpose way (without requiring argparse args for example).

These commits are currently in that PR, but may make sense to merge first, since #99 could take a while.

### DIFF
--- a/manubot/process/util.py
+++ b/manubot/process/util.py
@@ -271,9 +271,12 @@ def get_citekeys_df(citekeys: list, citekey_aliases: dict = {}):
     - standard_citekey: detagged_citekey standardized
     - short_citekey: standard_citekey hashed to create a shortened citekey
     """
-    citekeys_df = pandas.DataFrame({"manuscript_citekey": list(citekeys)}).drop_duplicates()
-    citekeys_df['detagged_citekey'] = citekeys_df.manuscript_citekey.map(
-        lambda citekey: citekey_aliases.get(citekey, citekey))
+    citekeys_df = pandas.DataFrame(
+        {"manuscript_citekey": list(citekeys)}
+    ).drop_duplicates()
+    citekeys_df["detagged_citekey"] = citekeys_df.manuscript_citekey.map(
+        lambda citekey: citekey_aliases.get(citekey, citekey)
+    )
     for citation in citekeys_df.detagged_citekey:
         is_valid_citekey(citation, allow_raw=True)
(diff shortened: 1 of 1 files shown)


manubot/process/util.py
```diff
@@ -285,51 +307,50 @@ def get_citekeys_df(args, text):
...
+        citekey_aliases = dict(
+            zip(tag_df["manuscript_citekey"], tag_df["detagged_citekey"])
         )
     else:
-        citekeys_df["detagged_citekey"] = None
+        citekey_aliases = {}
         logging.info(
             f"missing {args.citation_tags_path} file: no citation tags (citekey aliases) set"
         )
-    citekeys_df.detagged_citekey.fillna(
```
agitter: Does removing this `fillna` change the final data frame?

manubot/process/util.py
```diff
@@ -285,51 +307,50 @@ def get_citekeys_df(args, text):
...
-    requests  # require `import requests` in case this is essential for monkey patching by requests_cache.
-    requests_cache.install_cache(args.requests_cache_path, include_get_headers=True)
-    cache = requests_cache.get_cache()
-    if args.clear_requests_cache:
-        logging.info("Clearing requests-cache")
-        requests_cache.clear()
-    logging.info(
-        f"requests-cache starting with {len(cache.responses)} cached responses"
-    )
+    Parameters
```
agitter: Could list the other two parameters here: `requests_cache_path` and `clear_requests_cache`

manubot/process/util.py
dhimmel: No the functionality has been moved to:

```python
    citekeys_df["detagged_citekey"] = citekeys_df.manuscript_citekey.map(
        lambda citekey: citekey_aliases.get(citekey, citekey)
    )
```

manubot/process/util.py
dhimmel: done in https://github.com/manubot/manubot/pull/190/commits/0d1c6d113f23db142b8dcb8af805613cc212e1a1

dhimmel:
> My understanding is that there aren't any behavior changes with this refactoring. Is that correct?

Yes there should be no behavior changes.

agitter:
Thanks for the clarifications and docstring updates.

PR #197, opened 2020-01-17
Use r-lib/actions/setup-pandoc for pandoc install

dhimmel:
https://github.com/r-lib/actions/tree/master/setup-pandoc
https://twitter.com/maxheld/status/1218242371428798465

Thanks @jimhester for creating this action! Really makes using pandoc as part of the workflow much more straightforward!


dhimmel:
It looks like our windows builds [are not able](https://github.com/manubot/manubot/pull/197/checks?check_run_id=395839655#step:6:18) to detect the installed pandoc.

PR #208, opened 2020-02-19
Move build files from rootstock to new build submodule

dhimmel:
In preparation for manubot/template, we'd like to migrate as much functionality from Rootstock to this package. This PR adds the entire rootstock commit history and selectively takes some files using the approach in https://bneijt.nl/blog/post/merge-a-subdirectory-of-another-repository-with-git/.

One benefit of taking all commits is that it helps provide contributor credit to rootstock contributors even if we end up deprecating that repository. We could prune the history for only the files we're taking but that rewrites commit hashes and could be more clutter if in the future we pull other files from rootstock.


dhimmel:
The commands I used:

```shell
git fetch rootstock
git merge --strategy ours --no-commit --allow-unrelated rootstock/master
git read-tree --prefix=manubot/build/pandoc-data -u rootstock/master:build/assets
git read-tree --prefix=manubot/build/pandoc-data/defaults -u rootstock/master:build/pandoc-defaults
git read-tree --prefix=manubot/build/pandoc-data/themes -u rootstock/master:build/themes
git read-tree --prefix=manubot/build/pandoc-data/plugins -u rootstock/master:build/plugins
git commit
```

dhimmel:
Here's a rought PR (into this PR) to prototype calling pandoc https://github.com/dhimmel/manubot/pull/1

agitter:
I would also err on the side of making sure we don't erase contributor recognition, so this approach makes sense to me. I can review https://github.com/dhimmel/manubot/pull/1 once it's ready.  The design looks good so far.

dhimmel:
I am going to merge this PR into a new `build-refactor` branch, and then we can make PRs to this repo to continue with feature development.

PR #209, opened 2020-02-21
manubot build: run pandoc to render outputs

dhimmel:
Supersedes https://github.com/dhimmel/manubot/pull/1

### DIFF
--- a/manubot/build/build_command.py
+++ b/manubot/build/build_command.py
@@ -1,6 +1,7 @@
 import pathlib
 import logging
 import subprocess
+import os
 
 from manubot.util import shlex_join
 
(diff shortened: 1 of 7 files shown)


manubot/build/build_command.py
```diff
@@ -0,0 +1,119 @@
...
+        "--resource-path",
+        f".:{args.pandoc_data_dir.absolute()}",
+        "--defaults=common",
+        "--defaults=html",
+    ]
+    process = subprocess.run(command, cwd=args.directory)
+    logging.info(shlex_join(process.args))
+
+
+def build_pdf(args):
```
agitter: Are you planning to drop WeasyPrint support, or is that temporary?

PR #210, opened 2020-02-21
manubot process: require --skip-citations

dhimmel:
Remove support for citation processing in manubot process, in favor of using the pandoc-manubot-cite filter for all processing.

`pandoc-manubot-cite` does have a stricter syntax for what characters are allowed in citekeys. So we'll have to update the rootstock USAGE when upgrading manubot in Roostock.

My plan is to start more aggressively removing deprecated features like this to make the transition to `manubot/template` easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ Under this setup, you can run the Manubot using:
 
 ```sh
 manubot process \
+  --skip-citations \
   --content-directory=content \
   --output-directory=output
 ```
@@ -96,7 +97,7 @@ See `manubot process --help` for documentation of all command line arguments:
 usage: manubot process [-h] --content-directory CONTENT_DIRECTORY
                        --output-directory OUTPUT_DIRECTORY
                        [--template-variables-path TEMPLATE_VARIABLES_PATH]
-                       [--skip-citations] [--cache-directory CACHE_DIRECTORY]
+                       --skip-citations [--cache-directory CACHE_DIRECTORY]
                        [--clear-requests-cache]
                        [--log-level {DEBUG,INFO,WARNING,ERROR,CRITICAL}]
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -120,8 +121,10 @@ optional arguments:
                         `--template-variables-path=namespace=path_or_url`.
                         Namespaces must match the regex `[a-zA-
                         Z_][a-zA-Z0-9_]*`.
-  --skip-citations      Skip citation and reference processing. Specify when
-                        using the pandoc-manubot-cite filter. If citation-
+  --skip-citations      Skip citation and reference processing. Support for
```
agitter: Are you keeping this as a required argument to help differentiate calls to the old and current `manubot process`?

manubot/process/manuscript.py
```diff
@@ -3,24 +3,6 @@
 import json
 import logging
 import pathlib
-import re
-
-from manubot.cite.citekey import citekey_pattern, is_valid_citekey
-
-
-def get_citekeys(text):
```
agitter: I'm not reviewing the removed functions carefully.  The test cases should be sufficient.

manubot/process/manuscript.py
```diff
@@ -37,41 +19,11 @@ def get_text(directory):
...
-
-    if citekeys_df is not None:
-        # Number of distinct references by type
-        ref_counts = (
-            citekeys_df.standard_citekey.drop_duplicates()
-            .map(lambda x: x.split(":")[0])
-            .pipe(collections.Counter)
-        )
-        ref_counts["total"] = sum(ref_counts.values())
-        stats["reference_counts"] = ref_counts
```
agitter: Deep review reports these statistics: https://greenelab.github.io/deep-review/#continuous-collaborative-manuscript-drafting

We'll have to edit that part of the manuscript when we upgrade.

manubot/process/tests/test_manuscript.py
```diff
@@ -1,40 +1,5 @@
...
-    See https://github.com/manubot/manubot/issues/9
-    """
-    string_to_id = {
-        "url:https://github.com/manubot/manubot": "mNMayr3f",
-        "url:https://github.com/greenelab/manubot-rootstock": "1B7Y2HVtw",
-    }
-    text = "our new Manubot tool [@url:https://github.com/greenelab/manubot-rootstock; @url:https://github.com/manubot/manubot] for automating manuscript generation."
-    text = update_manuscript_citekeys(text, string_to_id)
-    assert "[@1B7Y2HVtw; @mNMayr3f]" in text
+def test_datetime_now():
```
agitter: Is this a placeholder?  Do you not want to remove the entire test file?

README.md
dhimmel: Yes, the idea is if you are using the old behavior of `manubot process` without `--skip-citations` you will get an error. So this is a way to ensure users are aware of the change and upgrade their workflow.

PR #214, opened 2020-03-05
Add Custom Filters

cgreene:
Manubot uses jinja2 for templating, but I couldn't find a way to add a custom filter. I wanted a way to randomize author contributions within a tier. I can use `groupby` to get the tiers, but there is no default filter for jinja2 that supports random.shuffle.

Here I implement a custom filter in a file within manubot (`filters.py`). The filter is also added to a dictionary `MB_FILTERS`. Any filter within `MB_FILTERS` is then included in the jinja2 environment. I have tested this for authors on the [deep-review](http://github.com/greenelab/deep-review) locally and it works. Two subsequent builds are shown below with randomized author order for v1 authors.

Screen Shot 2020-03-05 at 1 45 33 PM

Screen Shot 2020-03-05 at 1 46 02 PM


manubot/filters.py
```diff
@@ -0,0 +1,15 @@
+from random import Random, shuffle
+
+
+def mb_random(value, seed=None):
+    shuffled = list(value)
+    if seed is not None:
+        Random(seed).shuffle(shuffled)
+    else:
+        shuffle(shuffled)
+    return shuffled
```
dhimmel:
```suggestion
    shuffled = list(value)
    kwargs = {} if seed is None {"seed": seed}:
    Random(**kwargs).shuffle(shuffled)
    return shuffled
```

manubot/process/util.py
```diff
@@ -33,6 +33,7 @@
     is_valid_citekey,
     standardize_citekey,
 )
+from manubot.filters import MB_FILTERS
```
dhimmel: we should probably move this module to `manubot.process.jinja`. Probably shouldn't be a top-level submodule.

manubot/filters.py
cgreene: oh - wow - this is really elegant and allows other args too!

dhimmel:
My main worry is that this approach changes the author order only in the document frontmatter. However, `manubot.authors` is looped through internally in manubot to create the HTML head and set pandoc metadata:

https://github.com/manubot/manubot/blob/2ffcec1183bdaf7c49a81490b9c31d05e977b351/manubot/process/header-includes-template.html#L24-L37

https://github.com/manubot/manubot/blob/a022e715f266c0f116a6b38f993bd696dc62094a/manubot/process/util.py#L221

This could lead to an issue where the displayed authors mismatches the order of HTML `<head>` metadata (and metadata in other output formats).

And affiliation numbering will no longer be in order as seen above.

An alternative would be for manuscripts to shuffle `metadata.yaml` themselves in build.sh. This could be done with a small script that runs before `manubot process` on CI and outputs a new `metadata.yaml` with randomized authors.

This approach would not depend on jinja2, which has been a thorn in our sides!

@cgreene what do you think?

manubot/process/util.py
cgreene: I think that's a new module. If I understand correctly - done 👍

manubot/filters.py
cgreene: Actually Random takes only seed and not kwargs. I have a shorter implementation inbound.

dhimmel: Ah `*args`

cgreene:
I like the idea of shuffling metadata.yaml but it will require a lot of thought to get it right. Will the shuffle be implemented in manubot or is it something that can be implemented in a rootstock-derived repo? I think having this in the rootstock-derived repo is probably best.

To be honest, I also think implementing custom filters in core manubot is probably _usually_ the wrong solution. It would be nice to instead have them specified in the manuscript repo, and manubot would have some way to load them.

manubot/filters.py
cgreene: I'm not sure it makes sense to do args (although that's what I did the first time) because as far as I can tell the constructor takes no other types of args: https://docs.python.org/3/library/random.html#random.Random

dhimmel:
>  Will the shuffle be implemented in manubot or is it something that can be implemented in a rootstock-derived repo?

In a rootstock-derived repo by using a python script that reads metadata.yaml, shuffles authors, and then write metadata.yaml. You could use the same shuffle code.

>  I think having this in the rootstock-derived repo is probably best.

Agreed, in general I like these very manuscript-specific features to be demonstrated by a manuscript in the wild before adding to Manubot / Rootstock.

>  I also think implementing custom filters in core manubot is probably usually the wrong solution. It would be nice to instead have them specified in the manuscript repo, and manubot would have some way to load them.

Agree, I thought about defining jinja2 filters outside of the manubot package, since it would be more desirable, but didn't find an elegant solution.

dhimmel:
Now that we implemented a different approach in https://github.com/greenelab/deep-review/pull/997, should we close this PR?

I do think custom jinja2 filters defined at the manuscript-repo level is a good feature. Although, I want to delay it until needed in case we stop using jinja2 in the future.

cgreene:
Hmm - I would like manuscript-level custom filters as well. This plan sounds good to me. If there is a solution, it will look quite a bit different from this. Thanks!

cgreene:
I won't delete the branch just in case someone wants to take a stab at it before I get the chance to.

Issue #218, opened 2020-04-01
Support identifier.org compact identifiers as citation keys


dhimmel:
_Originally posted by @dhimmel in https://github.com/manubot/manubot/issues/216#issuecomment-607332692_:

> But perhaps the best general solution would be accept any identifiers.org compact IDs (CURIEs). Then:
> 
> 1. make API call to https://resolver.api.identifiers.org/clinicaltrials:NCT04292899 to extract `compactIdentifierResolvedUrl`, in this case https://clinicaltrials.gov/ct2/show/NCT04292899.
> 2. Use translation-server to generate bibliographic metadata for the URL.

dhimmel:
Here's an analysis of the current citation sources we support as per:

https://github.com/manubot/manubot/blob/db4c4771b30b08ce19209edfb1d0f819fef07a8b/manubot/cite/citekey.py#L20-L28

Are they supported at https://registry.identifiers.org/:

- `doi`: yes
- `pmid`: as `pubmed`
- `pmcid`: yes as `pmc`
- `arxiv`: yes
- `isbn`: yes
- `wikidata`: yes
- `url`: no, but we may want to migrate to `http` / `https` (also not supported)

For future citation key sources with special Manubot support, we probably should try to use the same prefix as identifiers.org when possible. We could deal with the special cases of `pmid` and `pmcid` pretty easily.

cthoyt:
I'm using manubot for the first time and I've got lots of ChEBI references in my first paper. Would love to see this feature!

However, a step further than what's possible with identifiers.org is that I would like to just use the identifier, and have the name get resolved and inserted in the manuscript. Then, we'd be a step closer to *not* having to do NER as much

dhimmel:
> I've got lots of ChEBI references in my first paper. Would love to see this feature!

Do you mean you'd like to cite specific compounds like `@CHEBI:53438` that would be expanded to cite https://www.ebi.ac.uk/chebi/searchId.do?chebiId=CHEBI:53438?

> I would like to just use the identifier, and have the name get resolved and inserted in the manuscript

Are you saying you'd like to just do `@53438` or `CHEBI:53438` or something else? Also do you want to create a citation & reference, or you'd like a way to use persistent IDs without a citation like:

> we used CHEBI:53438 to do the experiment

and have the rendered manuscript show

> we used [iron(3+) sulfate](https://www.ebi.ac.uk/chebi/searchId.do?chebiId=CHEBI:53438) to do the experiment

[...]

PR #219, opened 2020-04-15
cite.curie module for compact identifiers via identifiers.org

dhimmel:
refs https://github.com/manubot/manubot/issues/218
refs https://github.com/manubot/manubot/issues/220

### DIFF
--- a/manubot/cite/curie/__init__.py
+++ b/manubot/cite/curie/__init__.py
@@ -19,7 +19,7 @@ import json
 import logging
 import pathlib
 import re
-from typing import List, Tuple
+from typing import List, Set
 
 import requests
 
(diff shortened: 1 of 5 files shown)


dhimmel:
Currently stuck regarding how to deal with prefix capitalization as per https://github.com/identifiers-org/identifiers-org.github.io/issues/100

manubot/cite/curie/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+
+import requests
+
+_keep_namespace_fields = {
+    "prefix",
+    "mirId",  # MIRIAM Registry identifier
+    "name",
+    "pattern",  # regex pattern
+    "description",
+    "sampleId",  # example identifier
```
cthoyt, line 32: Another key you should keep is `namespaceEmbeddedInLui`. This signifies namespaces whose pattern starts with the prefix. For example, the GO pattern is `^GO:\d{7}$`, but to get a proper resolution you should only write it once like https://identifiers.org/GO:0006915 and *not like* https://identifiers.org/GO:GO:0006915

manubot/cite/curie/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+    # do not yet understand capitlaization
+    # https://github.com/identifiers-org/identifiers-org.github.io/issues/100
+    prefix_lower = prefix.lower()
+    if prefix_lower not in get_prefixes():
+        raise ValueError(
+            f"prefix {prefix_lower} for {curie} is not a recognized prefix"
+        )
+    resolver_url = "https://identifiers.org"
+    return f"{resolver_url}/curie"
```
cthoyt, line 101: missed squigglys?

manubot/cite/curie/__init__.py
dhimmel: Thanks. Excluding `namespaceEmbeddedInLui` was a crucial oversight. Unfortunately, even with knowledge of `namespaceEmbeddedInLui`, I'm worried it's insufficient to resolve all CURIEs with case-insensitive prefixes as per my comments in:

- https://github.com/identifiers-org/identifiers-org.github.io/issues/100#issuecomment-614679142
- https://github.com/identifiers-org/identifiers-org.github.io/issues/99#issuecomment-614690283

dhimmel:
@cthoyt you [mentioned](https://github.com/identifiers-org/identifiers-org.github.io/issues/99#issuecomment-614333476) you've done some related work at https://github.com/pyobo/pyobo. Thanks a lot for the feedback! Would definitely love to chat. Will reach out when I think your expertise is most valuable

dhimmel:
I'm reducing the scope of this PR. The new scope is just to download all identifiers.org namespaces and to expose a `manubot.cite.curie.curie_to_url` function. Future PRs will support citing CURIEs. This PR doesn't affect any existing manubot functionality.

Will address any additional review comments subsequently... merging now.

cthoyt:
@dhimmel i still owe you some updates to the ooh-na-na interface to include identifiers.org urls when applicable. sorry it's been slow getting to that

dhimmel:
> i still owe you some updates to the ooh-na-na interface to include identifiers.org urls when applicable.

no worries. Still a good amount of development until we're ready to use it!

agitter:
> Will address any additional review comments subsequently

I don't have any comments.  Looks like a good base for future CURIE citations and workarounds for the upstream issues.

PR #221, opened 2020-04-16
metadata: convert author.funders to list

dhimmel:
refs https://github.com/manubot/rootstock/issues/329

In local output, got the following for legacy affiliations.

```
## WARNING
/home/dhimmel/Documents/repos/manubot/manubot/process/util.py:150: DeprecationWarning: Expected list for Daniel S. Himmelstein's affiliations. Assuming multiple affiliations are `; ` separated. Please switch affiliations to a list.
  category=DeprecationWarning,
```


dhimmel:
I don't have any warnings when funders is a string and not a list. In this case, funders gets converted to `[funders]` and would thus appear as provided in the metadata.

PR #232, opened 2020-05-24
reduce travis and appveyor CI duties

dhimmel:
now that GitHub Actions is comprehensively testing and deploying,
use travis and appveyor only for tests where they provide unique
coverage.

Should help avoid intermittent failures from upstream resources
and avoid performing extra compute for little benefit.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -25,4 +25,4 @@ before_test:
   - ps: $env:Path += ";C:\Program Files\Pandoc\"
 
 test_script:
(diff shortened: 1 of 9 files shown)


dhimmel:
@agitter anything else you think we should continue testing for on appveyor / travis?

agitter:
These integration tests are a better idea than the single test I proposed.  I'm rerunning failed builds.  It looks like they were caused by intermittent arXiv and timestamp issues.

Issue #240, opened 2020-08-12
manubot cite json output order does not match order of input identifiers


vincerubinetti:
On Windows, I'm running the following command:
```
manubot cite --output research-2015.json \
  doi:10.1093/hmg/ddv440 \
  doi:10.1186/s12864-015-2250-5 \
  doi:10.1038/ncomms9019 \
  doi:10.1186/s13040-015-0050-8 \
  doi:10.1038/ng.3259 \
  doi:10.1038/nmeth.3249 \
  doi:10.1371/journal.pcbi.1004005 \
  pmc:PMC4299935 \
```
and the order of the json data that comes back is this:
```
10.1038/ncomms9019
10.1038/ng.3259
10.1038/nmeth.3249 
10.1093/hmg/ddv440
10.1186/s12864-015-2250-5
10.1186/s13040-015-0050-8
10.1371/journal.pcbi.1004005
PMC4299935
```
It looks like the results are returned in alphabetical order of their identifier? But perhaps it should match the order they are provided in. This would make programming with the command easier, avoiding additional step of matching the output back to the order of the input.

agitter:
It looks like this happens because `Citations.get_csl_items` calls `group_citekeys_by`:
https://github.com/manubot/manubot/blob/a57ccf0be6972329ff3010eaaa0c5df7ccebb2d5/manubot/cite/citations.py#L56-L68

The `sorted` call is there because `groupby` expects the iterable to already be sorted on the key.  To maintain the original order, I think we may have to explicitly store the original order and resort after `groupby`.

vincerubinetti:
Not a huge issue, but would make life easier in certain cases. At the very least, it should probably be documented in the readme.

Here's my use case, fwiw:

I'm redoing the [greenelab website](https://greenelab.github.io/greenelab.com/). For the list of publications we had, maintainers were previously just manually formatting and pasting in the citations. In the new version, we're using `manubot cite` to automatically get all the meta data like title, author, etc.

However, we want to sort the publications by date, and the templating language we're using to display the `manubot cite` json output is not good at sorting by deeply nested fields in a json object, like `issued: { "date-parts: [ [2020,11, 23] ] }`. So we were relying on the order we specified the identifiers.

dhimmel:
Yeah, the old implementation returned the references in the same order as the input. Let me look into the easiest way to restore the old behavior.

[...]

PR #241, opened 2020-08-14
cite command: preserve order of input citekeys in output

dhimmel:
closes https://github.com/manubot/manubot/issues/240

### DIFF
--- a/manubot/cite/citations.py
+++ b/manubot/cite/citations.py
@@ -25,6 +25,10 @@ class Citations:
     csl_item_failure_log_level: tp.Union[str, int] = "WARNING"
     # whether to prune csl items according to the JSON Schema
     prune_csl_items: bool = True
+    # whether to sort csl items by standard_id
+    # `sort_csl_items=False` retains order of input_ids in get_csl_items.
+    # (input_ids with the same standard_id will still be deduplicated).
+    sort_csl_items: bool = True
 
     def __post_init__(self):
         input_ids = list(dict.fromkeys(self.input_ids))  # deduplicate
@@ -54,7 +58,7 @@ class Citations:
         return remove
 
     def group_citekeys_by(
-        self, attribute: str = "standard_id"
+        self, attribute: str = "standard_id", sort: bool = True,
     ) -> tp.List[tp.Tuple[str, list]]:
         """
         Group `self.citekeys` by `attribute`.
@@ -63,9 +67,14 @@ class Citations:
         def get_key(x):
             return getattr(x, attribute)
 
-        citekeys = sorted(self.citekeys, key=get_key)
-        groups = itertools.groupby(citekeys, get_key)
-        return [(key, list(group)) for key, group in groups]
+        key_to_indices = dict()
+        for i, citekey in enumerate(self.citekeys):
+            key = get_key(citekey)
+            key_to_indices.setdefault(key, list()).append(i)
+        items = list(key_to_indices.items())
+        if sort:
+            items.sort(key=lambda item: item[0])
+        return [(key, [self.citekeys[i] for i in indices]) for key, indices in items]
 
     def unique_citekeys_by(self, attribute: str = "standard_id") -> list:
         return [citekeys[0] for key, citekeys in self.group_citekeys_by(attribute)]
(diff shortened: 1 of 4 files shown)


manubot/cite/citations.py
```diff
@@ -25,6 +25,10 @@ class Citations:
     csl_item_failure_log_level: tp.Union[str, int] = "WARNING"
     # whether to prune csl items according to the JSON Schema
     prune_csl_items: bool = True
+    # whether to sort csl items by standard_id
+    # `sort_csl_items=False` retains order of input_ids in get_csl_items.
+    # (input_ids with the same standard_id will still be deduplicated).
```
dhimmel, line 30: I decided against outputting duplicate CSL Items when multiple citekeys with the same standard_id are provided. We could do this for the CSL JSON output, but when using --render, Pandoc will not print the same reference twice. So for consistency, I thought it was best to deduplicate.

dhimmel:
Some annoying test failures that are unrelated. Ready for review

manubot/cite/citations.py
vincerubinetti: I don't think it's a good idea in general, but if that's the upstream behavior I guess there's not much you can do about it.

For my case, I'll have to either 1) make a strong note to the user that they must avoid duplicates or risk screwing up the metadata of the entire list, 2) dedupe the list myself, 3) make a significant effort to match input to output ids, 4) make an individual manubot call for each reference. 1) is dangerous, 3) is problematic and error prone. 2) might also be error prone, because I have to make sure to match the way you are deduping (what if a doi and a arxvir resolve to the same paper, does that count as a dupe). Therefore I'll probably go with 4).

How would you feel about adding an additional `dont-dedupe` option

dhimmel: > what if a doi and a arxvir resolve to the same paper, does that count as a dupe

not currently. arxiv and doi references won't be considered duplicates. This really only applies to citations with the same prefix, such as doi/DOI, pubmed/pmid, arxiv/arXiv.

> For my case, I'll have to either 1) make a strong note to the user that they must avoid duplicates or risk screwing up the metadata of the entire list

So your application depends on having the output the match the input on a one-to-one basis? Not just having the order preserved? I imagine you wouldn't want duplicates in the Greene Lab Publications Page?

There is another situation where the output length will be less than the input length: if metadata lookup fails for one of the citations or if the citation is faulty. In this case you will also get a nonzero exit code.

Another option is for you to write yoru won Python script that plugs directly into the manubot Citations API. This would give you fine-grained control. Or we could see if we could embed the input_ids for each csl_item using CSL JSON's new `custom` field, which could enable 3 more easily above.

vincerubinetti: I don't want duplicate papers on the Greene Lab page, no. But that would be a much softer, less critical error than the metadata from a different paper being applied to a paper.

I had assumed there would be some error code thrown if one of the citation lookups failed. That's okay with me because that should halt the python script, and the person running it would notice it and be forced to address the error. The order error is worse because it could go unnoticed.

I'll probably just go with the one-by-one call method because it seems the most foolproof. Otherwise, I have to rely closely on knowing and matching the implementation details of manubot cite.

FWIW, my thoughts are that pandoc made a mistake de-duping citations because it should be a "WYSIWYG" layout/conversion engine and not a logic engine. And I feel that manubot cite should return a one-to-one list of results, with an empty, null, or otherwise null-ish result for any citation it can't find in the list.

vincerubinetti: Thanks for addressing this so quickly!

dhimmel: > I had assumed there would be some error code thrown if one of the citation lookups failed

An error status (nonzero exit code) is returned and the CSL JSON that can be generated is returned via stdout. The error messages go to stderr.

> FWIW, my thoughts are that pandoc made a mistake de-duping citations

The thing is that `manubot cite` and what pandoc are doing is creating a bibliography or references list. For this application, you do want deduplication. So we're in the slightly awkward position of having the command being called "cite" but the infrastructure designed for bibliography generation.

The python Citations object ends up storing a dictionary called `input_to_csl_id` that maps from the input_id to the csl_item_id:

https://github.com/manubot/manubot/blob/3d6d5d4cb79a8674630a5a1ddaafbe04bc08ad80/manubot/cite/citations.py#L153

It sounds like this is what you want access to. We could consider an option for `manubot cite` to return an input_id_to_csl_item object.

vincerubinetti: > you do want deduplication. 

Still disagree with that. If you use a converter program to go from docx -> pdf for example, you don't expect it to try to be "smart" and mess with your content. KISS. But that is off topic.

> We could consider an option for manubot cite to return an input_id_to_csl_item object.

Something like this would be helpful. Most helpful would be the exact, case-sensitive query string the user input, to make it dead simple to match. For now though I'll have to do the one-paper-per-query approach.

PR #245, opened 2020-09-18
Fix oai logging

jmsmkn:
`{arxiv_id}` and `{response_arxiv_id}` were not being replaced as the f-string identifier was missing.


agitter (approved):
Thanks for the fix @jmsmkn.  The test failures are unrelated to this change, so we'll address those separately.

PR #246, opened 2020-09-24
Fix failing cite filter and black tests

agitter:
Update authors in Papernot et al. reference

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -32,18 +32,18 @@ def pytest_addoption(parser):
 @pytest.fixture(scope="session", autouse=True)
 def use_requests_cache(request):
     """Use requests_cache for caching web API calls.
-       Enabled with --use-requests-cache flag:
+    Enabled with --use-requests-cache flag:
 
-          pytest --use-requests-cache
-          pytest --use-requests-cache --requests-cache-hours=3
-          pytest --use-requests-cache --requests-cache-hours=8760
+       pytest --use-requests-cache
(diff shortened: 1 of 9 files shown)


agitter:
I ran black 19.10b0 locally and the output was"68 files left unchanged."  However, black 20.8b1 (the version that runs in the CI tests), found these changes in 8 files.

conftest.py
```diff
@@ -32,18 +32,18 @@ def pytest_addoption(parser):
...
-       - By default the cache persists for 12 hours. Default can be overrriden
-         with --requests-cache-hours option.
-       - Cache does not work for invocations of manubot commands as subprocess.
+    Notes:
+    - By default the cache persists for 12 hours. Default can be overrriden
+      with --requests-cache-hours option.
+    - Cache does not work for invocations of manubot commands as subprocess.
 
-       Cache file location: ./.cache/requests-cache.sqlite
+    Cache file location: ./.cache/requests-cache.sqlite
```
dhimmel, line 46: From https://black.readthedocs.io/en/stable/change_log.html

> Black now reindents docstrings when reindenting code around it (https://github.com/psf/black/pull/1053)

dhimmel (approved):
not sure which black change is resulting in long lines being split.

@agitter do you want to either merge or rebase merge such that these commits are kept separate?

thanks for fixing that CI failure.

agitter:
@dhimmel yes, I'll do a regular merge to keep these commits separate

Issue #248, opened 2020-10-25
Python 3.9 compatibility and GitHub Actions builds


dhimmel:
Python 3.9 was [released](https://www.python.org/downloads/release/python-390/) on Oct. 5, 2020 ([changelog](https://www.python.org/downloads/release/python-390/))!

It'd be nice to add Python 3.9 to our CI builds. Currently, GitHub Actions is blocked by https://github.com/actions/virtual-environments/issues/1740. Probably won't work on this issue until GitHub Actions support for Python 3.9 is 100%.

Once we have testing, we might discover some incompatibilities or warnings that should be addressed.

PR #250, opened 2020-10-30
Support and test on Python 3.9

dhimmel:
refs https://github.com/manubot/manubot/issues/248

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,8 +9,8 @@ environment:
     - PYTHON: "C:\\Python36-x64"
       PYTHON_VERSION: "3.6.x"
       PYTHON_ARCH: "64"
-    - PYTHON: "C:\\Python38-x64"
-      PYTHON_VERSION: "3.8.x"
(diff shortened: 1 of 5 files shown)


dhimmel:
Looks like everything is working with Python 3.9. The test failures appear to all be due to rate limiting of our NCBI API calls.

.appveyor.yml
```diff
@@ -1,17 +1,10 @@
-environment:
-
-  image: Visual Studio 2017
+image: Visual Studio 2019
 
+environment:
   matrix:
     - PYTHON: "C:\\Python36"
-      PYTHON_VERSION: "3.6.x"
-      PYTHON_ARCH: "32"
```
dhimmel, line 8: I'm not sure these are neccessary. possibly `PYTHON_ARCH` is, but I don't think so. `PYTHON_VERSION` is redundant with `PYTHON`

agitter (approved):
Looks good to me.  All tests pass after the rate limiter update.

PR #251, opened 2020-10-31
Add typing to manubot.cite.pubmed

dhimmel:
I'd like to begin more comprehensively typing this package, with the eventual goal of running a static type checker like mypy. In the meantime, these types help as documentation.

### DIFF
--- a/manubot/cite/csl_item.py
+++ b/manubot/cite/csl_item.py
@@ -24,9 +24,10 @@ for csl_data that is JSON-serialized.
 """
 
 import copy
+import datetime
 import logging
 import re
-from typing import List, Optional
+from typing import List, Optional, Union
 
 from manubot.cite.citekey import CiteKey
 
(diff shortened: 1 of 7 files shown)


manubot/cite/csl_item.py
```diff
@@ -149,7 +150,11 @@ def clean(self, prune: bool = True):
             self.validate_against_schema()
         return self
 
-    def set_date(self, date, variable="issued"):
+    def set_date(
+        self,
+        date: Union[None, str, datetime.date, datetime.datetime],
+        variable="issued",
```
agitter, line 156: I didn't catch this before merging, but should this be `variable: str = "issued",` to match line 168?

dhimmel: yeah good catch.

dhimmel: added in https://github.com/manubot/manubot/pull/252

Issue #258, opened 2020-11-22
cache citations using (only) CSL JSON when using the pandoc filter alone


slippycheeze:
G'day.  I'm using the manubot pandoc citation filter without the rest of the tooling, as it best fits the target audience for the document – a business setting.  Overall it works wonderfully, and I'm very pleased with being able to cite resources including arbitrary webpages with minimal hassle.

## summary of the request
1. `pandoc-manubot-filter` should store all downloaded citations in a CSL JSON file.
2. `pandoc-manubot-filter` should read that same CSL JSON if it exists, but no error if it doesn't.
3. `pandoc-manubot-filter` should add new automatic citations that are not already present in the CSL JSON file.
4. `pandoc-manubot-filter` should not update citations that are present in the CSL JSON file.
     - ...or, at least, it should not update the "accessed on" date of the items, other updates might be OK.

Having this behind a flag would be fine.  This is the default behaviour of `pandoc-url2cite`, and works well (for me) – manubot has a much better range of citation identifiers, and is more robust, so I'd like to keep using it, if possible, than fall back to `pandoc-url2cite`.

Also, if you think this is a reasonable request, but don't want to implement it, I'm happy to submit an implementation.  I'd rather know it was acceptable in principal first, though.

## details of this request

When citing arbitrary webpages the "accessed on" date is important, because it is the only persistent identifier for which version of the content was used.  For accuracy is is also useful to be able to manually set that date, if I actually used a previous version of the page, or whatever.

The "accessed on" date for arbitrary webpage citations (eg: `[@https://example.org/]`) is generated when the data is fetched, a reasonable choice – but one that means I need to use the same citation data each time, rather than having manubot fetch it each time the document is built.

The `manubot-requests-cache-path` option lets me store that data in an sqlite database of binary blobs, and then reuses the same data each time.  This solves the problem of getting a stable date for when a website was accessed, but it comes with a few problems along the way:

1. the citation data is not human-readable or editable in any reasonable way.
2. wiping the cache means *all* the records are fetched again, updating the "accessed on" date for *all* webpages, and
3. the URL is hashed, making it very difficult to delete a single cached record, even if I wanted to do so.

I can also write a CSL JSON file containing the resolved references, using `manubot-output-bibliography`, which is human readable and editable, and is easy enough to remove a single citation from.  manubot doesn't read that data by default, though, so it isn't a solution unless manubot is also told to read it with the `bibliography` metadata setting, but:

[...]

PR #261, opened 2020-11-23
manubot-pandoc-cite manubot-bibliography-cache

dhimmel:
refs https://github.com/manubot/manubot/issues/258


dhimmel:
@slippycheeze this PR enables the metadata variable `manubot-bibliography-cache`. It defaults to None / null / no caching, since I'm not sure there is a default path that works well for all users. We can always revisit the default decision in the future and either add one or emit a warning if no cache is specified.

Also added support for YAML output (detected by `.yaml` or `.yml` path suffix). Still a bit experimental, but might be nice for human readability/editability.

manubot/cite/citations.py
```diff
@@ -177,17 +178,36 @@ def csl_json(self) -> str:
...
+            default_flow_style=False,
+            width=float("inf"),
+            allow_unicode=True,
+            sort_keys=False,
+        )
+
+    def write_csl_items(self, path: tp.Union[os.PathLike, str, None]) -> None:
+        """
+        Write CSL Items to a JSON or YAML file at `path`.
+        If path ends with a .yml or .yaml extension, write as CSL YAML.
```
agitter, line 200: I like this strategy for YAML support.  Continue using JSON by default and support YAML optionally based on extension.

agitter, line 200: I like this strategy for YAML support. Continue using JSON by default and support YAML optionally based on extension.

manubot/pandoc/cite_filter.py
```diff
@@ -154,19 +162,32 @@ def process_citations(doc: pf.Doc) -> None:
 
     - bibliography (use to define reference metadata manually)
     - citekey-aliases (use to define tags for cite-by-id citations)
+    - manubot-bibliography-cache:
```
agitter, line 165: To make sure I'm following the usage correctly, to enable the desired caching would a user provide the same path to `manubot-bibliography-cache` and `manubot-output-bibliography`?  Or they could use different paths, but populate `manubot-bibliography-cache` by taking `manubot-output-bibliography` as a starting point and manually deleting any entries that should not be cached?

manubot/pandoc/cite_filter.py
```diff
@@ -195,7 +216,8 @@ def process_citations(doc: pf.Doc) -> None:
         req_cache.close()
 
     citations.write_citekeys_tsv(path=doc.get_metadata("manubot-output-citekeys"))
-    citations.write_csl_json(path=doc.get_metadata("manubot-output-bibliography"))
+    citations.write_csl_items(path=doc.get_metadata("manubot-output-bibliography"))
+    citations.write_csl_items(path=doc.manubot["bibliography_cache"])
```
agitter, line 220: I may not be following the behavior here.  All of the citations are written to both the output bibliography and the bibliography cache, overwriting the existing bibliography cache?

manubot/pandoc/tests/test_cite_filter/input.md
```diff
@@ -11,6 +11,7 @@ references:
   container-title: Greene Laboratory
   location: University of Pennsylvania
   issued: {'date-parts': [[2018, 12, 19]]}
+# manubot-bibliography-cache: manubot/pandoc/tests/test_cite_filter/bibliography-cache.json
```
agitter, line 14: Why is this commented out?

manubot/util.py
```diff
@@ -142,3 +143,35 @@ def read_serialized_dict(path: str) -> dict:
         f"Expected data encoded by {path!r} to be a dictionary at the top-level. "
         f"Received {data.__class__.__name__!r} instead."
     )
+
+
+def _yaml_str_representer(dumper: "yaml.Dumper", data: str):
+    """
+    Use YAML literal block style for multiline strings.
```
agitter, line 150: How universal is this literal block style?

agitter (commented):
This is looking good to me.  My comments pertain to usage to make sure I understand how a user will interact with the cache and how that addresses the proposed behavior in #258.

How is the cache populated the first time?  Is the idea that the file will not exist initially so it will not impact the first run of the cite filter.  But then it will be written and available after that first run?

I'm leaning toward leaving the new bibliography cache off by default.  I haven't thought through how it will help or hurt documents with references that intermittently fail.  Without the cache, the failures may eventually resolve themselves once the remote data source is responsive.  However, with a cache could we end up caching and reusing the minimal metadata from a failed request (e.g. the URL only)?

Are there other design questions from #258 I should comment on here?

manubot/util.py
dhimmel: I think any parser that fully implements the YAML 1.1 spec should be able to parse this.

At https://yaml-multiline.info/, the format we specify for multiline strings, matches the options under block scalars:

- Block Scalar Style: Keep newlines (literal)
- Block Chomping: No newline at end (strip)

From what I can tell, this format is the best if we envision the YAML being tracked with git.

manubot/pandoc/tests/test_cite_filter/input.md
dhimmel: For this test, don't want to use a CSL-bibliography cache. Uses --bibliography instead.

manubot/pandoc/cite_filter.py
dhimmel: `manubot-output-bibliography` and `manubot-bibliography-cache` and both default to None. If both are specified, two bibliography files can be written. I don't think most users will specify both.

Will improve docs.

dhimmel:
> I haven't thought through how it will help or hurt documents with references that intermittently fail. Without the cache, the failures may eventually resolve themselves once the remote data source is responsive. However, with a cache could we end up caching and reusing the minimal metadata from a failed request (e.g. the URL only)?

Great point. The fallback URL reference methods aren't really something we want to be caching. Perhaps in the future, we could set a field on the CSL Item custom field like cache=False.

PR #262, opened 2020-11-24
retelimit API calls to Zotero translation server

dhimmel:
Might reduce the frequency of failures on CI.

If it seems to be working, I'll clean up code.

### DIFF
--- a/manubot/cite/zotero.py
+++ b/manubot/cite/zotero.py
@@ -10,6 +10,9 @@ https://github.com/manubot/manubot/issues/82.
 import json
 import logging
 from typing import Dict, Any, List
+import functools
+import os
+import warnings
 
 import requests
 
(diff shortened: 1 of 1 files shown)


manubot/cite/zotero.py
```diff
@@ -142,3 +150,21 @@ def search_or_web_query(identifier: str) -> ZoteroData:
     else:
         zotero_data = search_query(identifier)
     return zotero_data
+
+
+@functools.lru_cache()
+def _get_zotero_rate_limiter() -> "RateLimiter":
+    """
+    Rate limiter to cap NCBI E-utilities queries to <= 3 per second as per
```
agitter, line 158: This NCBI comment isn't applicable for the Zotero use case.

agitter (commented):
It looks like there are still a fair number of test case failures.

dhimmel:
> It looks like there are still a fair number of test case failures.

Yeah, given that it didn't clearly fix the errors, I'm going to close this PR.

PR #267, opened 2020-12-03
manubot cite: remove --render & cslyaml support

dhimmel:
--format=csljson or --format=cslyaml will output the CSL Data without using Pandoc. Other choices will render with Pandoc.

By removing --render, this change is backwards incompatible, but the error message will indicate that --render is not an available option:

```
usage: manubot [-h] [--version] {process,cite,webpage} ...
manubot: error: unrecognized arguments: --render
```

I think there's a benefit in removing an unnecessary option. The one benefit of `--render` is that it's more succinct than `--format=plain`.

Now that pandoc support `--to=csljson`, it seems like we should adopt this same convention. cslyaml support will be convenient. Note pandoc does not have a --to=cslyaml option.

@agitter any thoughts?


agitter:
I actually like the `--render` option and use it somewhat regularly.  Even if it isn't necessary, it might be nice to leave it as a sort of syntactic sugar.

If we do get rid of it, I could adapt.

dhimmel:
If we wanted to leave `--render`, what would we have it do? If --render, default --format to plain rather than --csljson.

Another option would be to create some flags (mutually exclusive with --format)

- --yml for --format=cslyaml
- --md for --format=markdown
- --txt for --format=plain

Those are the formats I tend to use. How about you?

agitter:
> If --render, default --format to plain rather than --csljson.

That makes sense to me.

Regarding the flags, would syntax then be `manubot cite --txt <citekey>`?  That could be a good alternative as well.  I primarily use plain and markdown formats with the cite command.

dhimmel:
> Regarding the flags, would syntax then be `manubot cite --txt <citekey>`

yes. That would be equivalent to `manubot cite --format=plain <citekey>`. We could also make the argument `--plain`.

README.md
```diff
@@ -176,9 +176,10 @@ Additional usage information is available from `manubot cite --help`:
 
 <!-- test codeblock contains output of `manubot cite --help` -->
 ```
-usage: manubot cite [-h] [--output OUTPUT] [--render]
-                    [--format {plain,markdown,docx,html,jats}] [--csl CSL]
-                    [--bibliography BIBLIOGRAPHY] [--allow-invalid-csl-data]
+usage: manubot cite [-h] [--output OUTPUT]
+                    [--format {csljson,cslyaml,plain,markdown,docx,html,jats} | --yml | --txt | --md]
```
dhimmel, line 180: @agitter we could also do --cslyaml --yaml --plain --markdown. Here I went with most succinct, but it might be a bit confusion to not have format exactly match the shorthand.

dhimmel:
Note that --md output with pandoc 2.11 includes HTML fragments, but this is unrelated to this PR. Posted upstream at https://github.com/jgm/pandoc/issues/6921

manubot/cite/tests/cite-command-rendered/references-markdown-2.11.1.1.md
```diff
@@ -1,5 +1,5 @@
-<span class="csl-left-margin">1. </span><span class="csl-right-inline">**Generalization of the Fermi Pseudopotential**<span class="csl-block">Trang T. Le, Zach Osman, D. K. Watson, Martin Dunn, B. A. McKinney</span>*arXiv* (2018-06-14) <https://arxiv.org/abs/1806.05726v1><span class="csl-block">DOI: [10.1088/1402-4896/ab0811](https://doi.org/10.1088/1402-4896/ab0811)</span></span>
+1. **Generalization of the Fermi Pseudopotential**Trang T. Le, Zach Osman, D. K. Watson, Martin Dunn, B. A. McKinney*arXiv* (2018-06-14) <https://arxiv.org/abs/1806.05726v1>DOI: [10.1088/1402-4896/ab0811](https://doi.org/10.1088/1402-4896/ab0811)
```
dhimmel, line 1: markdown output with pandoc 2.11.1.1 is missing line breaks.

dhimmel:
@agitter  I am going to merge, but if you have any comments about https://github.com/manubot/manubot/pull/267#discussion_r535748272, we still have time to change before a release.

README.md
agitter: The way you set it up works for me.  It is slightly inconsistent, but since we're using it as shorthand I like the shorter versions (e.g. `--md` instead of `--markdown`).

PR #271, opened 2020-12-09
Use setuptools_scm for package version

dhimmel:
https://github.com/pypa/setuptools_scm/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,9 @@ pip-wheel-metadata/
 build/
 dist/
 
+## setuptools_scm
+manubot/version.py
+
 # System specific files
 
 ## Linux
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 # Python
 __pycache__/
 *.egg-info/
+.eggs/
 pip-wheel-metadata/
 .ipynb_checkpoints
 .cache
--- a/README.md
+++ b/README.md
@@ -352,17 +352,16 @@ manubot process \
 [![PyPI](https://img.shields.io/pypi/v/manubot.svg?logo=PyPI&style=for-the-badge)](https://pypi.org/project/manubot/)
 
 This section is only relevant for project maintainers.
-Travis CI deployments are used to upload releases to [PyPI](https://pypi.org/project/manubot).
+GitHub Actions is used to deploy releases to [PyPI](https://pypi.org/project/manubot).
 
-To create a new release, bump the `__version__` in [`manubot/__init__.py`](manubot/__init__.py).
-Then, set the `TAG` and `OLD_TAG` environment variables:
+To create a new release,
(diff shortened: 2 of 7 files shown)


dhimmel:
```bash
# remove cached files
rm -rf build dist/ manubot.egg-info/
# build wheel
python -m build --wheel > build-wheel.log
# list files in wheel, ensure no tests
unzip -l dist/manubot-*.whl
```

dhimmel:
The main remaining error seems to be with `pytest manubot/tests/test_command.py`, [logs](https://github.com/manubot/manubot/pull/271/checks?check_run_id=1527741665)

```
E       AssertionError: assert 'vNA' == 'v0.1.dev1+g654cdfb'
E         - v0.1.dev1+g654cdfb
E         + vNA
```

The challenge is that our tests are inside the application code. Therefore, we don't have a situation where all tests are run against the installed version. Instead tests run against the local code from the repository. This means the setuptools_scm-generated version is not always accessible. 

A heavy solution would be to switch to a src layout as per [these docs](- https://docs.pytest.org/en/stable/goodpractices.html#choosing-a-test-layout-import-rules):

> In this situation, it is strongly suggested to use a src layout where application root package resides in a sub-directory of your root:

Switching [import-modes](https://docs.pytest.org/en/stable/pythonpath.html#import-modes) might work if we remove relative imports from tests. 

It'd be nice to simplify the release process by single-sourcing the version to the git tag, but possibly this is more work than its worth.

agitter:
So once this is working, the version number will always follow this [default versioning scheme](https://github.com/pypa/setuptools_scm/#default-versioning-scheme)?  That is, it will only be the simple semantic version when the HEAD is the latest tag.

dhimmel:
> That is, it will only be the simple semantic version when the HEAD is the latest tag.

Yes.

dhimmel:
@agitter what are your thoughts on whether we should use setuptools_scm or not? I'm undecided.

agitter:
I initially liked the idea of automating versioning and slightly reducing the manual maintenance needed.  However, it seems to add complexity that isn't really necessary.  I find the simple semantic versions to be more readable and intuitive out of habit.  They are less precise, but in practice we haven't had a problem with Manubot versioning.  We either refer to very specific versions via git commit or the released versions.

dhimmel:
> I find the simple semantic versions to be more readable and intuitive out of habit. They are less precise, but in practice we haven't had a problem with Manubot versioning. We either refer to very specific versions via git commit or the released versions.

Yeah, agreed. I think it might be nice to have `{tag}` for no_distance_and_clean and `{tag}-{revision_hash}` for everything else. But this is not yet configurable in a declarative way, see https://github.com/pypa/setuptools_scm/issues/419. So I think I'll close this PR and we might reopen in future.

PR #273, opened 2020-12-09
fix exclusion of tests in wheel distribution

dhimmel:
We included the following to exclude tests from our distributions:

https://github.com/manubot/manubot/blob/0c2e5b28809acd7c9ca76152146ddc24345d0781/setup.cfg#L71-L73

But that doesn't actually work. It only works to exclude a top-level `tests` directory from being detected by `find_packages`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -71,6 +71,7 @@ dev =
 [options.packages.find]
 exclude =
     tests
+    *.tests
 
 [options.entry_points]
 console_scripts =


dhimmel:
```bash
# remove cached files
rm -rf build dist/ manubot.egg-info/
# build wheel
python -m build --wheel
# list files in wheel, ensure no tests
unzip -l dist/manubot-*.whl
```

Issue #123, opened 2019-06-26
Make source prefixes of citation keys optional


nichtich:
Source prefixes (`url:`, `doi:`, `pmcid:`, `wikidata:`...) are needles: [citation-js](https://citation.js.org/) can automatically detect URL, DOI, PMID, Wikidata identifier... to route requests for bibliographic data to different sources. Just use regular expressions to guess the source.

agitter:
I'm interested to learn more about this and what sources citation-js supports.  I'm still reading through its documentation and https://peerj.com/preprints/27466/.  The [input formats tutorial](https://citation.js.org/api/tutorial-input_formats.html) indicates DOIs, doi.org URLs, and Wikidata entity ids are the supported standard identifiers.  Does citation-js support general URLs, PMIDs, etc?

The [demo](https://citation.js.org/) seems consistent with that:
`const data = await Cite.async('10.7287/peerj.preprints.27466v1')`
works perfectly without the DOI prefix.

Using a URL like
`const data = await Cite.async('https://peerj.com/preprints/27466/')`
returns `Error: This format is not supported or recognized`.  Same for the arXiv ids, PMIDs, and PMCIDs I tried.

doi.org URLs like `https://doi.org/10.7287/peerj.preprints.27466v1` do work.

dhimmel:
I see three main benefits of including source prefixes in citations:

1. unambiguous citations for humans: `@pmid:31233491` lets human users know that the citation is pointing a PubMed record. I imagine few users would be able to immediately know `@31233491` was a PubMed ID.

2. unambiguous citations for machines: I am not sure that no identifiers are shared between sources. Currently, we support the following sources (and hopefully more in the future):
    https://github.com/manubot/manubot/blob/025b7c78e96d37cb81388bf739411dfe82969780/manubot/cite/util.py#L7-L15

3. Sources can be used to perform more stringent error checking of identifiers. Tags / citations keys become more confusing / difficult to handle when it's not clear whether something is a key or a resolved identifier. For example, `@31233491` could be a tag or a PubMed ID. Maybe the user intended it to be a tag, but had a typo, and thus it's now treated as an erroneous PubMed ID.

The main benefits I see to allowing users to omit the source prefix are:

1. convenience
2. unified behavior with other tools that do not require the prefix, like [citation-js](https://citation.js.org/) or zotero's [translation-server](https://github.com/zotero/translation-server)

[...]

PR #277, opened 2020-12-23
Infer citekey prefixes

dhimmel:
closes https://github.com/manubot/manubot/issues/123


manubot/cite/handlers.py
```diff
@@ -25,6 +25,25 @@
     "manubot.cite.wikidata.Handler_Wikidata",
 ]
 
+_infer_prefix_patterns = {
+    ("doi", "accession_pattern"),
+    ("doi", "shortdoi_pattern"),
+    ("pmc", "accession_pattern"),
+    ("pubmed", "accession_pattern"),
```
dhimmel, line 32: Note what now happens with `manubot cite --md 1`

> 1. **Formate assay in body fluids: application in methanol poisoning.**
AB Makar, KE McMartin, M Palese, TR Tephly *Biochemical medicine* (1975-06) <https://www.ncbi.nlm.nih.gov/pubmed/1>
DOI: [10.1016/0006-2944(75)90147-7](https://doi.org/10.1016/0006-2944(75)90147-7) · PMID: [1](https://www.ncbi.nlm.nih.gov/pubmed/1)

The other patterns are unlikely to match user-defined keys. But since PubMed IDs can start at 1, this pattern will pick up user defined citekeys that employ a simple numbering scheme. This wouldn't be catastrophic, since these citekeys would have to have metadata provided in a bibliography. However both of the citekeys would get `pubmed:` prefixed while using the metadata from bibliography.

Another options would be to use a stricter regex for inferring a pubmed prefix. For example, six digit minimum.

References:

- https://phabricator.wikimedia.org/T154905

manubot/cite/handlers.py
```diff
@@ -25,6 +25,25 @@
     "manubot.cite.wikidata.Handler_Wikidata",
 ]
 
+_infer_prefix_patterns = {
+    ("doi", "accession_pattern"),
+    ("doi", "shortdoi_pattern"),
+    ("pmc", "accession_pattern"),
```
dhimmel, line 31: `pmc:pmc4304851` and  `pmc4304851` both fail since PMC is lowercase. But with the former you get a warning:

```
## WARNING
Inspection of dealiased citekeys revealed potential problems:
pmc:pmc4304851 -- PubMed Central Identifiers must start with 'PMC'.
```

manubot/cite/handlers.py
agitter: Because there are valid PMIDs that are shorter than the six digit minimum, then I think Manubot should support those the same as longer PMIDs.  Using numbering scheme for user defined citekeys isn't a great idea even if it is valid.

manubot/cite/handlers.py
```diff
@@ -25,6 +25,25 @@
     "manubot.cite.wikidata.Handler_Wikidata",
 ]
 
+_infer_prefix_patterns = {
```
agitter, line 28: Is there a reason this is a set instead of a list that specifies the precedence of patterns?  Currently all of the patterns match unique strings, right?  Will that always be true in the future?

manubot/pandoc/cite_filter.py
```diff
@@ -25,6 +25,11 @@
   Useful when a citation is used many times or contains invalid characters.
   Aliases can also be defined in markdown with link reference syntax.
 
+- manubot-infer-citekey-prefixes (boolean):
```
agitter, line 28:
```suggestion
- `manubot-infer-citekey-prefixes` (boolean):
```

manubot/pandoc/cite_filter.py
```diff
@@ -48,6 +53,10 @@
 - `manubot-clear-requests-cache` (boolean):
   If true, clear the requests cache at `manubot-requests-cache-path`.
 
+- `manubot-fail-on-errors` (boolean):
```
agitter, line 56: Is this adding documentation of an existing feature?

agitter (commented):
This looks good to me overall.  I'm glad that inferring prefixes can be optionally disabled in the pandoc-manubot-cite filter as well.

manubot/pandoc/cite_filter.py
dhimmel: yes I realized we didn't document this option

manubot/cite/handlers.py
dhimmel: mistake. fixed in https://github.com/manubot/manubot/pull/277/commits/fcf9a9421ef8d5a9cbc9dabdf95f07af2153ce60

Issue #290, opened 2021-06-03
No way to stop manubot from adding rootstock remote


twrightsman:
I use manubot in a subdirectory of a larger git repository and do not use the rootstock repository. It would be great if I could set a command-line flag or environment variable to tell manubot not to forcibly add the rootstock remote to my git repo. Would you be open to a pull request making this change? I currently have a patched manubot that does not add the remote but it would be nice if it were an upstream feature.

dhimmel:
Thanks for the issue. Do you have CI build your manubot manuscript? Or is up to the users to build the manuscript inside the repo subdirectory? Is your repo public? If so it might be helpful to see.

Here's where the `rootstock` remote gets added:

https://github.com/manubot/manubot/blob/3ff3000f76dcf82a30694d076a4da95326e3f6ae/manubot/process/metadata.py#L195-L202

This is needed to compute `rootstock_commit`, which AFAICT is a debugging field at the moment and gets written to the output [`variables.json`](https://github.com/manubot/rootstock/blob/3fb3a2af232c6da4107074831b72122518fbcdac/variables.json#L85).

I agree that `manubot process` always setting the remote is too aggressive. Another user had this issue in https://github.com/manubot/manubot/issues/229.

We could make this opt-in, where perhaps rootstock-derived repos would opt-in by default? I'm open to hear other solutions?

twrightsman:
Ah sorry, I didn't find that issue in my search before posting!
The repo is not public yet and I only build it manually, currently no plans for a CI build. It pretty much looks like the rootstock repository with a few files removed.
Reading the other issue, I can see why the remote needs to be added for CI purposes and it looks like `manubot build` will solve my use case when it's finished.
I think we can close this issue as a duplicate, since my use case of keeping related code with the manuscript has already been discussed in other issues.

agitter:
Because `manubot build` doesn't have a timeline, I like the idea of adding a new `manubot process` command line flag that could disable adding the rootstock remote.  By default the flag would be set so that the remote is added, but advanced users would have more control to support custom use cases.

@dhimmel would a small pull request to add such a flag make sense to you?

dhimmel:
> would a small pull request to add such a flag make sense to you?

Yes! Will reopen.

agitter:
@twrightsman you mentioned that you
> currently have a patched manubot that does not add the remote

Are you interested in initiating a pull request?  We'll be happy to review and edit it with you.

twrightsman:
[...]

PR #292, opened 2021-06-11
Add command-line argument to skip adding the rootstock git remote

twrightsman:
closes https://github.com/manubot/manubot/issues/290

Addresses https://github.com/manubot/manubot/issues/290 and related issues by adding the `--skip-remote` command-line flag to `manubot process` that will skip over fetching the rootstock commit ID and therefore not adding the rootstock git remote to the local git repository.

This makes life a bit easier for those using manubot in a non-standard way, such as outside the rootstock repository.

Implementation is up for discussion but I've tested this one and it works locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ usage: manubot process [-h] --content-directory CONTENT_DIRECTORY
                        --output-directory OUTPUT_DIRECTORY
                        [--template-variables-path TEMPLATE_VARIABLES_PATH]
                        --skip-citations [--cache-directory CACHE_DIRECTORY]
-                       [--clear-requests-cache]
+                       [--clear-requests-cache] [--skip-remote]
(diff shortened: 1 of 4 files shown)


dhimmel:
Thanks a lot for the PR. Can you run the following:

https://github.com/manubot/manubot/blob/3ff3000f76dcf82a30694d076a4da95326e3f6ae/manubot/tests/test_readme.py#L53-L54

dhimmel:
Also see the formatting changes required by black:

https://github.com/manubot/manubot/pull/292/checks?check_run_id=2810181010#step:4:91

twrightsman:
@dhimmel Are these test failures expected? They look unrelated.

dhimmel (approved):
The test errors are not expected, but rather we rely on a bunch of flaky upstream APIs. So these test failures happen frequently and are unrelated to this PR.

agitter:
Thanks for the new feature @twrightsman

PR #304, opened 2021-10-29
Upgrade identifiers.org database on 2021-10-29

dhimmel:


dhimmel:
`test_web_query_returns_single_result_pubmed_url` failed due to PubMed being currently down. So will merge despite failure.

Issue #305, opened 2021-10-29
Switch from Identifiers.org to Bioregistry


dhimmel:
@cthoyt [mentionted](https://twitter.com/cthoyt/status/1442522899970154498) [Bioregistry](https://bioregistry.io/) as something that could be useful for Manubot.

Currently we use identifiers.org to define a list of CURIE prefixes that we can resolve, when used as citation keys. Bioregistry is more comprehensive (it is a superset of Identifiers.org), has an open contribution model on [GitHub](https://github.com/biopragmatics/bioregistry), and fixes a lot of issues.

There are two possible implementations:

1. The first is to replace the data in [`manubot/cite/curie/namespaces.json`](https://github.com/manubot/manubot/blob/11f7192e32aa280145e118135f0041a41a1b02d1/manubot/cite/curie/namespaces.json) with Bioregistry data. But [`bioregistry.json`](https://raw.githubusercontent.com/biopragmatics/bioregistry/main/src/bioregistry/data/bioregistry.json) has a different schema. @cthoyt is there a bioregistry data file with just the bioregistry-approved fields (like the top-level fields seen at https://bioregistry.io/api/registry/3dmet?format=yaml)?

2. using the `bioregistry` python package. This would require more work to implement but we could remove a lot of lines of code in favor of an upstream project! One concern is that there are [some dependencies](https://github.com/biopragmatics/bioregistry/blob/9278ec189f05d3b1d5669cf3127ba49988f54c26/setup.cfg#L47-L61).

cthoyt:
I'll look back into making a slimmer version of the Bioregistry python package's requirements since some are only needed for updating the data

cthoyt:
You might want to check the TSV export in https://github.com/biopragmatics/bioregistry/blob/main/docs/_dat/registry.tsv

cthoyt:
As of https://github.com/biopragmatics/bioregistry/pull/235, there's now a nice JSON and YAML export with the consensus registry

cthoyt:
I did #238 to remove some of the requirements from base install from the Bioregistry. Now the ones that remain are all pretty lightweight. I'm not sure if the issue was just that there were a lot of requirements, or from a different perspective. The ones that are left all have minimal requirements themselves (if any at all)

Issue #282, opened 2021-01-24
Get metadata directly from bioRxiv


cthoyt:
It would be nice to be able to use `biorxiv:` as a citation prefix similarly to using `doi:`. I might have missed it, but I think the preferred way is to use a URL and the following code parses out the bioRxiv identifier, then formats it into a DOI

https://github.com/manubot/manubot/blob/553090a16ac24a9fdc3f8df241a91b790c140865/manubot/cite/citekey.py#L306-L313

Rather than generating a DOI for a bioRxiv identifier by parsing it out of a URL based on a pattern (as the following code appears to be doing), it could even be possible to query the bioRxiv API directly. I recently made a [pull request](https://github.com/SuLab/WikidataIntegrator/pull/169) to the Wikidata Integrator project to hit the bioRxiv API and generate Wikidata items. The same logic could be adapted to generate CSL items in Manubot. This might provide a solution to #16 as well.

dhimmel:
Regarding the proposed `biorxiv:accession` identifiers, here's what I'm imagining:

- `biorxiv:269860`
- `biorxiv:269860v2` (for versioned citation)
- `biorxiv:2020.08.20.259226` (newer bioRxiv URLs / DOIs include date). Would `biorxiv:259226` work?
- `biorxiv:2020.08.20.259226v2`
- what about `medrxiv:2020.10.29.20222208v2`?

> I might have missed it, but I think the preferred way is to use a URL and the following code parses out the bioRxiv identifier, then formats it into a DOI

The function `url_to_citekey` is never called when processing a manuscript. It's only used by other applications that directly call the API. Currently there are three ways to cite a biorxiv preprint:

1. by DOI like `doi:10.1101/2020.08.20.259226`. Metadata retrieved by manubot using DOI content negotiation.
2. by bioRxiv URL like `https://www.biorxiv.org/content/10.1101/2020.08.20.259226v2`. Uses Zotero translator.
3. by DOI URL like `https://doi.org/10.1101/2020.08.20.259226`.  Uses Zotero translator.

For 2 and 3 (and all URLs currently), metadata retrieval is delegated to Zotero [translation-server](https://github.com/zotero/translation-server) which uses [these javascript translators](https://github.com/zotero/translators). Currently it looks like Zotero detects the metadata from bioRxiv pages using the embedded metadata (HTML `<meta>` tags). It also recognizes the DOI in the page, but probably defaults to the embedded metadata.

I see two distinct issues:

[...]

PR #223, opened 2020-05-12
Citation refactor: CiteKey & Handle classes, CURIE support

dhimmel:
merges https://github.com/manubot/manubot/pull/223
supersedes https://github.com/manubot/manubot/pull/147
closes https://github.com/manubot/manubot/issues/218
refs https://github.com/manubot/manubot/issues/192

Enables citing any CURIE registered with identifiers.org
Preferred prefixes have been updated as per identifiers.org:

- pmid -> pubmed
- pmcid -> pmc

Legacy prefixes are still supported but get converted to
standard prefixes during standardization.

Bare http and https prefixes are now supported.
Prefixes are no longer case sensitive, although the standard form is lowercase.

Create a CiteKey class for representing citekeys and their various
stages of IDs. Create a Handler class to handle citekeys based on
their prefixes.

Additional citation infrastructure should be migrated to the citekey class,
but probably in subsequent PRs.

PR #306, opened 2021-11-03
CURIE support: switch to Bioregistry

dhimmel:
refs https://github.com/manubot/manubot/issues/305


manubot/cite/handlers.py
```diff
@@ -419,207 +667,360 @@ def get_csl_item(self, citekey) -> Dict[str, Any]:
...
+    "icd-11": "manubot.cite.curie.Handler_CURIE",
+    "icd-o": "manubot.cite.curie.Handler_CURIE",
+    "icd10": "manubot.cite.curie.Handler_CURIE",
+    "icd10cm": "manubot.cite.curie.Handler_CURIE",
+    "icd11": "manubot.cite.curie.Handler_CURIE",
+    "icd9": "manubot.cite.curie.Handler_CURIE",
+    "icd9_2005": "manubot.cite.curie.Handler_CURIE",
+    "icd9cm": "manubot.cite.curie.Handler_CURIE",
+    "icd9cm_2005": "manubot.cite.curie.Handler_CURIE",
+    "icd9cm_2006": "manubot.cite.curie.Handler_CURIE",
```
dhimmel, line 781: Actual ICD support makes me happy

manubot/cite/handlers.py
```diff
@@ -419,207 +667,360 @@ def get_csl_item(self, citekey) -> Dict[str, Any]:
...
     "interpro": "manubot.cite.curie.Handler_CURIE",
+    "ip": "manubot.cite.curie.Handler_CURIE",
+    "ipd-imgt/hla": "manubot.cite.curie.Handler_CURIE",
+    "ipr": "manubot.cite.curie.Handler_CURIE",
     "ird.segment": "manubot.cite.curie.Handler_CURIE",
     "irefweb": "manubot.cite.curie.Handler_CURIE",
+    "iro": "manubot.cite.curie.Handler_CURIE",
     "isbn": "manubot.cite.isbn.Handler_ISBN",
+    "isbn-10": "manubot.cite.curie.Handler_CURIE",
+    "isbn-13": "manubot.cite.curie.Handler_CURIE",
```
dhimmel, line 833: we might want to use `manubot.cite.isbn.Handler_ISBN` for `isbn-10` and `isbn-13`

manubot/cite/curie/__init__.py
```diff
@@ -177,28 +147,29 @@ def standardize_curie(curie):
...
 def curie_to_url(curie):
     """
     `curie` should be in `prefix:accession` format
     """
     curie = standardize_curie(curie)
-    resolver_url = "https://identifiers.org"
-    return f"{resolver_url}/{curie}"
+    prefix, accession = curie.split(":", 1)
+    registry = get_prefix_to_registry()[prefix.lower()]
+    if "url" in registry:
```
cthoyt, line 168:
```suggestion
if "uri_format" in registry:
    return registry["uri_format"].replace("$1", accession)
```

manubot/cite/tests/test_curie.py
```diff
@@ -1,69 +1,67 @@
...
-            )
-        if namespace["prefix"] != namespace["curiePrefix"].lower():
-            print(
-                f"{namespace['prefix']} identifiers use "
-                f"curiePrefix {namespace['curiePrefix']}"
-            )
+    bioregistry = get_bioregistry(compile_patterns=True)
+    assert isinstance(bioregistry, list)
+    reports = list()
+    for registry in bioregistry:
```
cthoyt, line 15: I would call each entry in the bioregistry a resource, not a registry.

cthoyt:
Earlier today I did a lot of work to make the requirements for the bioregistry lighter. Maybe you could reconsider just reusing the code from the bioregistry package itself to do a lot of the things in here

manubot/cite/tests/test_curie.py
dhimmel: Ah i came to this conclusion independently and updated in https://github.com/manubot/manubot/pull/306/commits/6db6ae6b47652086bc4407d6cca8ee3ba1c677b4. Nice to see that we picked the same term "resource"

cthoyt: Here's the glossary of all of the terms I'm using: https://cthoyt.com/2021/10/07/biopragmatics-glossary.html

Will port it to a more neutral location that isn't my blog in the near future

manubot/cite/curie/__init__.py
dhimmel: So the next daily build that writes to https://github.com/biopragmatics/bioregistry/blob/main/exports/registry/registry.json will change `url` to `uri_format`?

manubot/cite/tests/test_curie.py
dhimmel: What about that slide you showed with definitions of things like thesuarus and vocabulary... is that somewhere online?

manubot/cite/curie/__init__.py
cthoyt: It should have happened already, you've helped me catch a typo! New version coming momentarily ;)

then of course it will update properly on the nightly build

dhimmel:
> Earlier today I did a lot of work to make the requirements for the bioregistry lighter. Maybe you could reconsider just reusing the code from the bioregistry package itself to do a lot of the things in here

Yeah. This PR is possibly just an intermediate step. It has already simplified things, like not having to deal with embedded namespaces and expanding to resource-URLs rather than identifiers.org URLs (which has pros and cons).

Need to think more about what calling `bioregistry` functions directly would look like in the context of the `Handler` system here.

BTW, should we filter deprecated prefixes or perhaps warn the user if they use any deprecated prefixes?

manubot/cite/tests/test_curie.py
cthoyt: https://bit.ly/WPCI2021

I will be putting together a summary of the whole workshop by the end of this week hopefully

cthoyt:
> Yeah. This PR is possibly just an intermediate step. It has already simplified things, like not having to deal with embedded namespaces and expanding to resource-URLs rather than identifiers.org URLs (which has pros and cons).

🚀 

> Need to think more about what calling `bioregistry` functions directly would look like in the context of the `Handler` system here.

I could send a demo of replacing some of the simple bits

> BTW, should we filter deprecated prefixes or perhaps warn the user if they use any deprecated prefixes?

Until now I haven't really taken a hard stance on deprecated resources. If it's possible to resolve them, then I guess it's still fine, but maybe helpful to the user to know they're using something outdated. However, if manubot can't also suggest a better alternative, then this might just be annoying

manubot/cite/curie/__init__.py
cthoyt: Done as of https://github.com/biopragmatics/bioregistry/pull/244

manubot/cite/tests/test_citekey.py
```diff
@@ -76,7 +76,7 @@
         pytest.param(
             "DOID:14330",
             dict(
-                standard_id="doid:14330",
+                standard_id="DOID:14330",
```
dhimmel, line 79: There's one change in this PR I'd like to get some feedback on before merge @cthoyt and @agitter. In the past, all of our standardized citation keys used lowercase prefixes. However, this PR switches the prefix capitalization to the `preferred_prefix` in Bioregistry (for citation keys handled using `CURIE_Handler`).

Our past standard prefixes were all lowercases, as noted in https://github.com/manubot/manubot/pull/223:

> Prefixes are no longer case sensitive, although the standard form is lowercase.

But it's not clear this is a policy / requirement or just a statement that was true for all the prefixes with existing support.

The benefit of using `preferred_prefix` is that it will match the authoritative capitalization from the resource creator. For example, the Disease Ontology uses `DOID:14330` and not `doid:14330`. On the other hand, it adds complexity versus all lower prefixes and is perhaps less aesthetic (subjective).

I think the switch to preferred prefix makes sense, but wanted to get some additional feedback. Users are free to create citations with any capitalization of the prefix, but this is what the input citations get standardized to.

cthoyt: I considered the same thing for the bioregistry. I believe case insensitivity and the ability to use synonyms is much more practical from the user's perspective, and then the bioregistry could be used to standardize it internally before rendering a citation.

dhimmel: > I believe case insensitivity and the ability to use synonyms is much more practical from the user's perspective

FTR Manubot will (and does) support case insensitive prefixes in citation keys as inputs as well as synonymous prefixes. The question here is whether standard CURIEs should use preferred prefix capitalization or lowercase prefixes. Bioregistry currently uses lowercase, but that seemed to be a sticking point for some resource creators.

cthoyt: oh then I would definitely suggest using the preferred prefix when available.

manubot/cite/handlers.py
agitter: Yes, that makes sense to process all ISBNs consistently

manubot/cite/tests/test_citekey.py
agitter: I don't have a strong opinion on this but agree with using preferred prefix capitalization.

agitter:
> BTW, should we filter deprecated prefixes or perhaps warn the user if they use any deprecated prefixes?

What happens when a prefix is deprecated?  If Manubot can no longer resolve it, then some sort of intervention seems necessary.  Would that act like a typical URL citation to a broken URL?

If the prefix is deprecated and cannot be resolved, I still lean toward providing a warning.

manubot/cite/handlers.py
```diff
@@ -222,83 +263,180 @@ def get_csl_item(self, citekey) -> Dict[str, Any]:
     "bionumbers": "manubot.cite.curie.Handler_CURIE",
     "bioportal": "manubot.cite.curie.Handler_CURIE",
     "bioproject": "manubot.cite.curie.Handler_CURIE",
+    "bioregistry": "manubot.cite.curie.Handler_CURIE",
+    "bioregistry.collection": "manubot.cite.curie.Handler_CURIE",
+    "bioregistry.schema": "manubot.cite.curie.Handler_CURIE",
+    "biorxiv": "manubot.cite.curie.Handler_CURIE",
```
agitter, line 269: This is relevant to #282.

Is it right that support for the `chemrxiv` prefix is not being added here?

manubot/cite/handlers.py
```diff
@@ -358,35 +539,86 @@ def get_csl_item(self, citekey) -> Dict[str, Any]:
...
+    "enm": "manubot.cite.curie.Handler_CURIE",
     "ensembl": "manubot.cite.curie.Handler_CURIE",
     "ensembl.bacteria": "manubot.cite.curie.Handler_CURIE",
     "ensembl.fungi": "manubot.cite.curie.Handler_CURIE",
     "ensembl.metazoa": "manubot.cite.curie.Handler_CURIE",
     "ensembl.plant": "manubot.cite.curie.Handler_CURIE",
     "ensembl.protist": "manubot.cite.curie.Handler_CURIE",
+    "ensemblglossary": "manubot.cite.curie.Handler_CURIE",
+    "entrez": "manubot.cite.curie.Handler_CURIE",
+    "entrez gene/locuslink": "manubot.cite.curie.Handler_CURIE",
```
agitter, line 566: Will whitespace and punctuation in a prefix cause problems anywhere for Manubot?

cthoyt: IMO these should be disallowed in Manubot - this will probably cause a lot of havoc with CURIE parsing. These synonyms are in the Bioregistry because other resources are propagating this kind of thing and it provides extended abilities for parsing malformed CURIEs, but Manubot probably shouldn't get into this business nor enable people to do this kind of things

manubot/cite/handlers.py
cthoyt: Related: all of the chemrxiv stuff is broken since they switched from figshare to their new closed system

cthoyt:
> > BTW, should we filter deprecated prefixes or perhaps warn the user if they use any deprecated prefixes?
> 
> What happens when a prefix is deprecated? If Manubot can no longer resolve it, then some sort of intervention seems necessary. Would that act like a typical URL citation to a broken URL?
> 
> If the prefix is deprecated and cannot be resolved, I still lean toward providing a warning.

Interestingly, being deprecated and being resolvable are pretty independent. I think Manubot should yell at users (or even disallow it completely) if they use a prefix that can't be resolved, but not necessarily make a big fuss for deprecated resources. That being said, most deprecated resources can't be resolved.

manubot/cite/handlers.py
dhimmel: It looks like `chemrxiv` is not prefix in bioregistry, so that seems correct for it not to be added here.

manubot/cite/handlers.py
dhimmel: Okay so should we filter prefix + synonyms to those matching the proposed prefex regex at https://github.com/biopragmatics/bioregistry/issues/158

> New prefixes must validate against the following regular expression: ^[a-z][a-z0-9]+(\.[a-z][a-z0-9]+?)$

So manubot would only support the subset of prefixes that are not malformed?

dhimmel:
> I think Manubot should yell at users (or even disallow it completely) if they use a prefix that can't be resolved

How do we tell if a prefix is resolvable? Whether the resource has a `uri_format` field?

cthoyt:
> > I think Manubot should yell at users (or even disallow it completely) if they use a prefix that can't be resolved
> 
> How do we tell if a prefix is resolvable? Whether the resource has a `uri_format` field?

Yes, exactly. The `uri_format` should also not generate opaque URLs, either (e.g., tons of resources create ad-hoc identifiers.org URLs, but that doesn't actually mean they go somewhere, so this should be avoided)

manubot/cite/curie/__init__.py
```diff
@@ -104,18 +104,22 @@ def _download_bioregistry():
...
+        resource["all_prefixes"] = sorted(
             {
                 prefix,
-                *(x.lower() for x in metadata.get("synonyms", [])),
+                *(x.lower() for x in resource.pop("synonyms", [])),
             }
         )
-    registry = list(results.values())
+        registry.append(resource)
     json_text = json.dumps(registry, indent=2, ensure_ascii=False)
```
cthoyt, line 123: why not just do `sort_keys=True` in `json.dumps` to get more determinism?

dhimmel: The non-determinism was with values of `synonyms`, so `sort_keys` would not fix it. Also python dicts are now deterministic so sort_keys=False doesn't introduce non-determinism in field order.

PR #307, opened 2021-11-07
Refactor CURIE handling with Bioregistry package

cthoyt:
This PR is a proof-of-concept that some of the functionalities of CURIE management introduced in #306 could be covered by the Bioregistry. A big caveat of this implementation is that it removes some of the flexibility in handling of CURIEs, but this could also be a potential benefit since improvements meant for Manubot could be just directly PR'd/incorporated into the Bioregistry

### DIFF
--- a/manubot/cite/curie/__init__.py
+++ b/manubot/cite/curie/__init__.py
@@ -30,6 +30,7 @@ References:
 - [On the road to robust data citation](https://doi.org/10.1038/sdata.2018.95)
 - [Uniform Resolution of Compact Identifiers for Biomedical Data](https://doi.org/10.1038/sdata.2018.29)
 """
+
 import dataclasses
 import functools
 import json
@@ -72,7 +73,7 @@ class Handler_CURIE(Handler):
         try:
             self.resource = get_prefix_to_resource()[self.prefix_lower]
         except KeyError:
-tgith            raise ValueError(f"Unrecognized CURIE prefix {self.prefix_lower}")
+            raise ValueError(f"Unrecognized CURIE prefix {self.prefix_lower}")
         self.standard_prefix = (
             self.resource.get("preferred_prefix") or self.resource["prefix"]
         )
--- a/manubot/cite/curie/__init__.py
+++ b/manubot/cite/curie/__init__.py
@@ -32,13 +32,12 @@ References:
 """
 
 import dataclasses
-import functools
 import json
 import pathlib
 import re
 import typing
 
-import bioregistry
+from bioregistry import Manager, Resource
 
 from manubot.cite.citekey import CiteKey
 from manubot.cite.handlers import Handler
@@ -53,9 +52,16 @@ _keep_bioregistry_fields = {
     "synonyms",
 }
 
-
 bioregistry_path = pathlib.Path(__file__).parent.joinpath("bioregistry.json")
 
+
+def _get_manager() -> Manager:
+    with bioregistry_path.open() as file:
+        resources = json.load(file)
+    return Manager({entry["prefix"]: Resource(**entry) for entry in resources})
+
+
+manager = _get_manager()
 
 valid_prefix_pattern = re.compile(r"^[a-z0-9][a-z0-9._-]+?$")
 """
@@ -70,16 +76,10 @@ of a pandoc citation key without requiring escaping.
 @dataclasses.dataclass
 class Handler_CURIE(Handler):
     def __post_init__(self):
-        try:
-            self.resource = get_prefix_to_resource()[self.prefix_lower]
-        except KeyError:
+        self.resource = manager.get_resource(self.prefix_lower)
+        if self.resource is None:
             raise ValueError(f"Unrecognized CURIE prefix {self.prefix_lower}")
-        self.standard_prefix = (
-            self.resource.get("preferred_prefix") or self.resource["prefix"]
-        )
-        self.prefixes = self.resource["all_prefixes"]
-        if "pattern" in self.resource:
-            self.accession_pattern = self.resource["pattern"]
+        self.standard_prefix = self.resource.get_preferred_prefix()
 
     def get_csl_item(self, citekey: CiteKey):
         from ..url import get_url_csl_item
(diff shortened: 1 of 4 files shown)


manubot/cite/curie/__init__.py
```diff
@@ -67,16 +76,13 @@
...
 class Handler_CURIE(Handler):
     def __post_init__(self):
-        try:
-            self.resource = get_prefix_to_resource()[self.prefix_lower]
-        except KeyError:
+        self.resource = manager.get_resource(self.prefix_lower)
+        if self.resource is None:
             raise ValueError(f"Unrecognized CURIE prefix {self.prefix_lower}")
-        self.standard_prefix = (
-            self.resource.get("preferred_prefix") or self.resource["prefix"]
```
dhimmel, line 75: Shouldn't we still do `self.standard_prefix = self.resource.get_preferred_prefix()` here, since we want the standard prefix to have the preferred capitalization?

manubot/cite/curie/__init__.py
```diff
@@ -129,64 +134,33 @@ def _download_bioregistry() -> None:
...
-        )
-    try:
-        prefix, accession = curie.split(":", 1)
-    except ValueError:
-        raise ValueError(
-            f"curie must be splittable by `:` and formatted like `prefix:accession`. Received {curie}"
-        )
-    handler = Handler_CURIE(prefix.lower())
-    return f"{handler.standard_prefix}:{accession}"
+    return manager.normalize_curie(curie)
```
dhimmel, line 151: will this use preferred prefix capitalization?

manubot/cite/handlers.py
```diff
@@ -157,6 +157,252 @@ def get_csl_item(self, citekey: CiteKey) -> Dict[str, Any]:
     "4dn": "manubot.cite.curie.Handler_CURIE",
     "4dn.biosource": "manubot.cite.curie.Handler_CURIE",
     "4dn.replicate": "manubot.cite.curie.Handler_CURIE",
+    "AAO": "manubot.cite.curie.Handler_CURIE",
```
dhimmel, line 160: IIRC the keys here should be all lowercase, because lookup of handlers is always done with lowercase prefixes. So these capitalized handlers should never get called.

manubot/cite/curie/__init__.py
```diff
@@ -129,64 +134,33 @@ def _download_bioregistry() -> None:
             *(x.lower() for x in resource.pop("synonyms", [])),
         }
         # remove invalid prefixes as per https://github.com/manubot/manubot/pull/306#discussion_r744125504
-        resource["all_prefixes"] = sorted(
+        resource["synonyms"] = sorted(
             filter(valid_prefix_pattern.fullmatch, all_prefixes)
         )
```
dhimmel, line 139: Noting that we're going to use the initial key here of `synonyms` to refer to the Manubot modified synonym list, which has been all lowercase, includes the primary prefix, and removes bad prefixes.

manubot/cite/curie/__init__.py
```diff
@@ -129,64 +134,33 @@ def _download_bioregistry() -> None:
...
         registry.append(resource)
     json_text = json.dumps(registry, indent=2, ensure_ascii=False)
     bioregistry_path.write_text(json_text + "\n", encoding="utf-8")
```
dhimmel, line 142: Noting that we're still downloading the bioregistry from `https://github.com/biopragmatics/bioregistry/raw/main/exports/registry/registry.json` (a hardcoded URL) rather than using the bioregistry package to retrieve it. This PR preserves the design where a specific version of this package will be hardcoded to a specific data version of the bioregistry. But presumably the bioregistry has the ability to download the bioregistry dataset or perhaps comes with a version of the dataset?

dhimmel (commented):
This is really interesting! Mostly just questions so I can better understand the effect of this PR. Thanks.

> A big caveat of this implementation is that it removes some of the flexibility in handling of CURIEs

How so? What flexibility is removed?

manubot/cite/curie/__init__.py
cthoyt: Yes it's also possible to completely switch, but doing it this way lets you decide when you want to update the bioregistry and potentially do your own overrides

manubot/cite/handlers.py
```diff
@@ -98,9 +98,9 @@ class Handler:
     """
 
     prefix_lower: str
-    prefixes = []
+    prefixes: List[str] = dataclasses.field(default_factory=list)
```
dhimmel, line 101: Cherry picked e8f9e8660f661b6b66b42f5ed8cd12a03fa2773d for https://github.com/manubot/manubot/pull/311 and noticed:

Setting the type annotation to `List[str]` causes the dataclass to consider this a field and causes subclasses to silently fail to set prefixes. For example, the following stops working:

https://github.com/manubot/manubot/blob/b0b0a1873a79dc5ab216fa731ba05817621e1938/manubot/cite/isbn.py#L12-L14

cthoyt: Ahh. Then it needs to be `ClassVar[List[str]]` but still I would advise against making the default as a list, since it is mutable and shared among all subclasses

dhimmel: Ah good to know about ClassVar. That is indeed what I'm looking for.

So the options are:

```py
# option 1
prefixes: ClassVar[List[str]] = []

# option 2
prefixes: ClassVar[List[str] | None] = None

# option 3
prefixes: ClassVar[List[str]] = dataclasses.field(default_factory=list)
```

Are any of these right? Option 1 has the shared mutability issue. Option 2 requires special casing to treat None as an empty list. Option 3 might be totally wrong because `dataclasses.field` is for instance variables not class variables, right?

cthoyt: Option 2 is the way to go

cthoyt: Just an extra note - you're right that option 3 is totally wrong. A fourth option would be just to type annotate `prefixes: ClassVar[List[str]]`, since you assume all subclasses _must_ provide this list. Unfortunately there isn't something like "abstract class variable" you can annotate on here, so your options for enforincing this are:

1. Duck typing (just assume it, and some code will fail if it's ever wrong)
2. Overkill write your own metaclass
3. Lightweight check in `__init_subclass__` that with hasattr on the class itself

Issue #309, opened 2022-01-07
request timeout when running pandoc-manubot-cite


phildong:
Hey first of all thanks for the great tool!
I have been happy with manubot until recently I'm getting a "502 bad gateway" issue for all of my citations.
Here is an example message:
```
## ERROR
Error fetching metadata for doi:10.1007/978-3-319-67558-9_33.
Invalid response from https://data.crosscite.org/10.1007/978-3-319-67558-9_33:
<html>
<head><title>502 Bad Gateway</title></head>
<body>
<center><h1>502 Bad Gateway</h1></center>
</body>
</html>
```

I'm not sure this is the correct place to report since it looks like a temporary crosscite.org website issue.
However I have been trying for a few days now and I can never successfully finish a build.
I'm wondering given the instability of crosscite.org what's the best way to get around this problem and get the manuscript properly built?
I do have a local cached `reference.json` generated by manubot automatically, I'm wondering if there's an easy way to tell manubot to just use that (without manually changing all my citation keys of course)
Thanks in advance for your help!

dhimmel:
We also got a similar error for a catalog [build](https://github.com/manubot/catalog/runs/4698320469?check_suite_focus=true#step:5:221):

```
ERROR: Error fetching metadata for doi:10.1101/474403.
Invalid response from https://data.crosscite.org/10.1101/474403:
<html>
<head><title>503 Service Temporarily Unavailable</title></head>
<body>
<center><h1>503 Service Temporarily Unavailable</h1></center>
</body>
</html>
```

Slightly different error code (502 versus 503). 

The 502 might be the same issue as https://github.com/crosscite/content-negotiation/issues/104.

I'll follow up with some fixes Manubot can implement, but in the meantime you can provide the metadata manually. Are you using Manubot to write a manuscript? If so, use [these instructions](https://github.com/manubot/rootstock/blob/c4a394e4293751fdca07f17bb731a54ebdfdb50c/USAGE.md#reference-metadata).

dhimmel:
We should consider reverting part of e9128567387d7e7eaa7b0a7765d8f06ca2238dc5 to switch the default for `content_negotiation_url` to `https://doi.org` from `https://data.crosscite.org`.


```sh
# this works
curl --silent --location \
  --header "Accept: application/vnd.citationstyles.csl+json" \
  https://doi.org/10.1007/978-3-319-67558-9_33

# this fails
curl --silent --location \
[...]

PR #310, opened 2022-01-07
get_doi_csl_item: try crosscite if datacite fails

dhimmel:
refs https://github.com/manubot/manubot/issues/309

The DataCite content negotiation (via `https://data.crosscite.org`) tends to produce better metadata but fails often. Adds fallback to CrossCite content negotiation (via `https://doi.org`).

### DIFF
--- a/manubot/cite/doi.py
+++ b/manubot/cite/doi.py
@@ -2,6 +2,7 @@ import json
 import logging
 import urllib.parse
 import urllib.request
+from functools import partial
 from typing import Any, Callable, Optional
 
 import requests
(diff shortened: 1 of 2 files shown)


manubot/cite/tests/test_doi.py
```diff
@@ -23,9 +25,13 @@ def test_expand_short_doi_not_short():
         expand_short_doi("10.1016/S0933-3657(96)00367-3")
 
 
-def test_get_doi_csl_item_crosscite():
+@pytest.mark.parametrize(
+    "get_doi_csl_item_negotiation",
```
agitter, line 29: Could it be confusing to have the argument name be the same as the manubot.cite.doi function name?

agitter (approved):
Looks good to me.  Consider my comment below optional.

manubot/cite/tests/test_doi.py
dhimmel: Perhaps, but get_doi_csl_item_datacite and get_doi_csl_item_crosscite are really just versions of get_doi_csl_item_negotiation with different default argument values.

manubot/cite/doi.py
```diff
@@ -146,6 +146,16 @@ def get_doi_csl_item_crosscite(doi: str):
...
 
 
+get_doi_csl_item_datacite = partial(
+    get_doi_csl_item_negotiation,
+    content_negotiation_url=content_negotiation_url_datacite,
+)
+get_doi_csl_item_crosscite = partial(
+    get_doi_csl_item_negotiation,
+    content_negotiation_url=content_negotiation_url_crosscite,
+)
```
dhimmel, line 156: `partial` introduces a bug in

https://github.com/manubot/manubot/blob/b8a00c0f33fe7eafed3515154a190f3aefd19182/manubot/cite/doi.py#L203-L212

https://github.com/manubot/manubot/runs/4748725137?check_suite_focus=true#step:6:1113

dhimmel: removing partial in https://github.com/manubot/manubot/pull/312

PR #312, opened 2022-01-08
get_doi_csl_item_negotiation: do not use partial

dhimmel:
fixes https://github.com/manubot/manubot/pull/310/files#r780685302

partial does not set `__name__` causing downstream error message to
raise an exception.

### DIFF
--- a/manubot/cite/doi.py
+++ b/manubot/cite/doi.py
@@ -2,7 +2,6 @@ import json
 import logging
 import urllib.parse
 import urllib.request
-from functools import partial
 from typing import Any, Callable, Optional
 
 import requests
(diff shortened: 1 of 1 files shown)


dhimmel:
CI [still fails](https://github.com/manubot/manubot/runs/4748969391?check_suite_focus=true#step:6:454) due to a datacite outage, but now we get the right error message:

```
ERROR    root:doi.py:142 Error fetching metadata for doi:10.1038/ng.3834.
Invalid response from https://data.crosscite.org/10.1038/ng.3834:
<html>
<head><title>502 Bad Gateway</title></head>
<body>
<center><h1>502 Bad Gateway</h1></center>
</body>
</html>
```

Issue #315, opened 2022-01-16
author order/annotation details


jessegmeyerlab:
It seems like some repositories can randomize author order and there also might be a way to define the author order without re-ordering the list in metadata.yml? I looked at https://github.com/manubot/rootstock/blob/main/USAGE.md and the issues in rootstock and here but couldn't find how to do this. Could this be added to the documentation?

jessegmeyerlab:
A related question is how to get the summary of words everyone contributed for determining author order? Sorry if I missed these points in the documentation

dhimmel:
Hello! I'm guessing this is in relation to https://github.com/jessegmeyerlab/proteomics-tutorial. Was very excited by the announcement on [twitter](https://twitter.com/J_my_sci/status/1482045053849661445).

> there also might be a way to define the author order without re-ordering the list in metadata.yml

currently, authors are read via this [source code](https://github.com/manubot/manubot/blob/7a2a2df1a803420a15d9aeac712c6640a8b6b388/manubot/process/util.py#L193-L206). The order is determined solely by the order in the `authors` field.

> some repositories can randomize author order

See https://github.com/greenelab/deep-review/pull/997. It's not in the docs since randomizing authors was viewed as a rather custom operation.

Perhaps we could add an actual manubot option, such that you could put `manubot-randomize-author-order: true` in the `metadata.yaml` file.

If you want to brainstorm a bit more about your ideal author order behavior, we can decide whether a Manubot option or manuscript-specific solution makes more sense.

> summary of words everyone contributed for determining author order

There's some prototype code in [this notebook](https://github.com/greenelab/meta-review/blob/2482af490804891c7f00b248c4e7ef7f356a0730/analyses/deep-review-contrib/01.contrib-git.ipynb), which created the data for [this figure](https://greenelab.github.io/meta-review/v/2482af490804891c7f00b248c4e7ef7f356a0730/#fig:contrib). Note that it's based on commits to the main branch such that the counts don't represent the actual words in the final product.

Personally, I don't think word count contributed should be the only factor in author order, since it doesn't capture all types of contributions and not all words are equal, especially since these are words/characters in the markdown source (imagine words added in a large HTML comment in the source). But don't want to dissuade you from generating these metrics if you find them helpful.

jessegmeyerlab:
Thanks Daniel! 

[...]

PR #317, opened 2022-01-19
Support metadata.manubot-randomize-author-order

dhimmel:
refs https://github.com/manubot/manubot/issues/315

Enables setting the following in metadata.yaml:

```yaml
manubot-randomize-author-order: true
```

### DIFF
--- a/manubot/process/metadata.py
+++ b/manubot/process/metadata.py
@@ -85,6 +85,23 @@ def _thumbnail_path_to_url(path):
     return url
 
 
+def get_head_commit() -> Optional[str]:
+    from manubot.util import shlex_join
+
+    args = ["git", "rev-parse", "HEAD"]
+    try:
+        return subprocess.check_output(
+            args, stderr=subprocess.PIPE, universal_newlines=True
+        ).strip()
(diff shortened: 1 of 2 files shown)


dhimmel:
Doesn't include an [`--only-on-ci`](https://github.com/greenelab/deep-review/blob/44ff95a2756bd03a32c046eb8983ac9c4f223b0a/build/randomize-authors.py#L24-L28) option for the randomization. Not sure that is really important. @agitter do you agree?

agitter:
I agree we can skip the `--only-on-ci` option.

Do we want to support a manual seed?  One workflow might be to randomize authors in the early stages of writing and then fix one final randomized order when the manuscript is submitted or published.

dhimmel:
> Do we want to support a manual seed

added in https://github.com/manubot/manubot/pull/317/commits/62f786fa5f7ce321c41a4ec18304aff7e1833cd0 to allow setting `manubot-randomize-author-order-seed`.

agitter (approved):
It looks like the test failures are due to the known DOI issues.

PR #318, opened 2022-01-19
Drop Python 3.6 support, add Python 3.10 support

dhimmel:
See https://discuss.python.org/t/python-3-6-rides-into-the-sunset/12964:

> Python 3.6 which officially reached end-of-life on 2021-12-23


dhimmel:
Hitting the following issue when [testing](https://github.com/manubot/manubot/runs/4862155587?check_suite_focus=true#step:6:451) in `pytest manubot/webpage/tests/test_webpage_command.py::test_webpage_command`:

```
OpenTimestamp command returned nonzero code (-6).
>>> ots stamp webpage/v/testing/index.html
```

Only happening on macOS, not linux or windows.

agitter:
@dhimmel should we update this line when we change supported Python versions?
https://github.com/manubot/manubot/blob/90925eeff90e3aaacaa504fc9a47107609cffc17/pyproject.toml#L10

dhimmel:
> should we update this line when we change supported Python versions?

Yes. Done in https://github.com/manubot/manubot/commit/3551decae866dea6bfc548033469babfec0c6c31.

PR #334, opened 2022-04-11
Add sec to filtered handlers for pandoc-secnos

matthewturk:
This filters out references to `sec:whatever` so that [pandoc-secnos](https://github.com/tomduck/pandoc-secnos) can be used.

### DIFF
--- a/manubot/cite/handlers.py
+++ b/manubot/cite/handlers.py
@@ -10,9 +10,9 @@ from .citekey import CiteKey
 
 """
 Non-citation prefixes used by the pandoc-xnos suite of Pandoc filters,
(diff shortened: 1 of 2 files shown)


agitter (commented):
This looks reasonable to me.  @dhimmel do you have any thoughts?

We've considered moving away from pandoc-xnos, but that is unlikely to happen in the short term.

dhimmel (approved):
Thanks @matthewturk! Looks good to me, especially since

1. The `sec` prefix is [not registered](https://bioregistry.io/registry/sec) with bioregistry
2. pandoc-crossref, another pandoc filter for references, also uses `sec:` for [section labels](https://lierdakil.github.io/pandoc-crossref/#section-labels)

Issue #337, opened 2022-06-15
Allow user to set timeout.


xihh87:
Users should be able to set the timeout for the requests to allow a more responsive manubot.

dhimmel:
Noting the `requests` docs on the timeout parameter: https://requests.readthedocs.io/en/latest/user/advanced/#timeouts

By the way, was this motivated by a specific URL that was failing to respond in a timely manner? And do you remember what that URL was?

xihh87:
This was motivated because I tried a manubot cite command that took over 10 minutes to search for 5 isbn.

```
$ time manubot cite --csl https://raw.githubusercontent.com/citation-style-language/styles/master/apa.csl --txt isbn:9789587784251 isbn:9798665944401 isbn:9781090802965 isbn:978-1491910399 isbn:9780128029145
## WARNING
Error parsing search_query output as JSON for isbn:9798665944401:
No items returned from any translator
## WARNING
Error in get_isbn_csl_item_zotero for 9798665944401 due to a JSONDecodeError:
[Errno Expecting value] No items returned from any translator: 0
## WARNING
Error in get_isbn_csl_item_citoid for 9798665944401 due to a KeyError:
'Metadata for ISBN 9798665944401 not found at https://en.wikipedia.org/api/rest_v1/data/citation/mediawiki/9798665944401'
Benjamin Smith. (2020). Ciencia De Datos - Guía completa para principiantes aprende sobre los reinos de la ciencia de datos de la A a la Z.

Garc??a Herrero, J., Molina L??pez, J. M., Berlanga de Jes??s, A., Patricio Guisado, M. ??ngel., Bustamante, ??lvaro. L., Padilla R, & Washington. (2018). Ciencia de datos: t??cnicas anal??ticas y aprendizaje estad??stico. Un enfoque pr??ctico. Alfaomega.

Jones, H. (2019). Ciencia de los datos: la gu??a definitiva sobre an??lisis de datos, miner??a de datos, almacenamiento de datos, visualizaci??n de datos, Big Data para empresas y apriendizaje autom??tico para principiantes.

Talia, D., Trunfio, P., & Marozzo, F. (2016). Data Analysis in the Cloud: Models, Techniques and Applications. http://www.sciencedirect.com/science/book/9780128028810

Wickham, H., & Grolemund, G. (2016). R for data science: import, tidy, transform, visualize, and model data (First edition). O’Reilly.

real	11m39.976s
user	0m0.705s
sys	0m0.070s
```

[...]

Issue #332, opened 2022-04-05
Quarto integration


agitter:
Quarto describes itself as "an open-source scientific and technical publishing system built on Pandoc":
- [website](https://quarto.org/)
- [GitHub](https://github.com/quarto-dev/quarto-cli)

I've only briefly looked through its features.  It uses standard Pandoc [citations](https://quarto.org/docs/authoring/footnotes-and-citations.html), so it may be able to benefit from Manubot's citation by identifier.

dhimmel:
Yes, I think a quarto extension to enable manubot's citation by identifier could be popular. Haven't really assessed the feasibility of this but here are some rudimentary links:

- https://quarto.org/docs/extensions/
- https://github.com/quarto-ext

It also might be interesting to see if Manubot could be built on top of quarto, where manubot would become mostly the extra tooling needed for:

- citation by identifier
- continuous integration and deployment
- git based versioning

agitter:
A basic question (that may be answered in the docs I haven't read carefully enough) is whether extensions are pandoc filters or something quarto specific.  [Creating filters](https://quarto.org/docs/extensions/filters.html) leads me to believe they are pandoc filters.  One next step could be to take some existing quarto document and see what happens if we try to run the Manubot filter on it.

cderv:
Hello. I am available if you have any question related to how Quarto works. 

Thanks @VladimirAlexiev for the answer. We do use Pandoc's Lua filter with Quarto internally but also to extend the feature available with some Extensions made by the community. It works with the Pandoc Lua API and also with some addition by the Quarto LUA API. (https://quarto.org/docs/extensions/lua-api.html)

> One next step could be to take some existing quarto document and see what happens if we try to run the Manubot filter on it

Did you try that ? Any filter that works with Pandoc should be able to be applied on a Quarto document. Maybe some ajustement would be needed if there are some conflicts between what you filter expect, and some processing that Quarto could have made before the filter is ran. 

Anyway, happy to help.

agitter:
I started exploring this with Quarto 1.2.313 and the example document in https://quarto.org/docs/get-started/authoring/text-editor.html. I created a new conda environment with Jupyter, the required Python packages, and manubot (which contains the `pandoc-manubot-cite` Pandoc filter).

Quarto/Pandoc seem to recognize `pandoc-manubot-cite` when I list it as one of the `filters` in the YAML block at the beginning of the document. Listing a gibberish name for a filter instead fails with an error
```
[...]

PR #338, opened 2022-06-15
WIP for #337 Allow setting timeout.

xihh87:


xihh87:
@agitter, @dhimmel

What do you think about this proposal?

xihh87:
@agitter is something wrong with this patch?

dhimmel:
Sorry about the slow reply. My bad.

I'm in favor of setting a timeout on requests that we've seen can get stuck, especially those that timeout rather than fail immediately for bad identifiers. Not sure if this is more than just the ISBN requests.

Don't really think we need to make the timeout configurable at the CLI level just yet nor do we need to pass timeout to everything that makes requests. Possibly we do want some global requests config, but I'm not sure this PR does that in a scalable way.

So in the interest of quickly merging, can we reduce this PR (or a new one) to just address the requests that are known to timeout?

xihh87:
I think giving good defaults and a mechanism to alllow the user control is the cleanest way to account for all unexpected edge cases:

- user has a 56kb modem, no problem, just increase timeout to 5 min.
- realtime editing (i.e. #332), just enable cahing and set a real low timeout…
- everything in between.

I also think it's cleaner to upgrade the interface that makes requests to handle this case and it would be more cumbersome to track where are all the special cases.

Now, I wold like to unify the network request interfaces to allow keeping track of the rate limit for APIs and caching but I'm counting on this changes [to speed up tests during development](https://github.com/manubot/manubot/issues/339#issuecomment-1161100827).

xihh87:
Those test fail because tests are not respecting rate limits for some APIs now that tests are quicker.

I would like that to be the next issue I work with.

PR #341, opened 2022-07-15
Fix test failing on main. 

xihh87:
It seems that this information was corrected on the queried source.

The comment states that this test should give another value,
so we use that value and the test passes again.

The order of the authors is unaffected and that seems to be the tested parameter.

Before this change is applied, this test fails on main branch.


dhimmel (approved):
Thank you for keeping the tests happy!

PR #342, opened 2022-07-15
#337 Set default timeout on ISBN search.

xihh87:
I've reduced the scope of changes on #338 [as requested](https://github.com/manubot/manubot/pull/338#issuecomment-1183912661):

> So in the interest of quickly merging, can we reduce this PR (or a new one) to just address the requests that are known to timeout?
>
> _Originally posted by @dhimmel in https://github.com/manubot/manubot/issues/338#issuecomment-1183912661_**

This change is only configurable on source code in the global variable defined on isbn search file.

### DIFF
--- a/manubot/cite/isbn.py
+++ b/manubot/cite/isbn.py
@@ -4,6 +4,8 @@ import re
 
 from .handlers import Handler
 
+default_timeout = 3
+
 
 class Handler_ISBN(Handler):
 
@@ -15,6 +17,9 @@ class Handler_ISBN(Handler):
 
     def inspect(self, citekey):
         import isbnlib
+
+        isbnlib.config.setthreadstimeout(seconds=timeout_seconds)
+        isbnlib.config.seturlopentimeout(seconds=timeout_seconds)
 
         fail = isbnlib.notisbn(citekey.accession, level="strict")
         if fail:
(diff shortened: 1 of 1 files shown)


manubot/cite/isbn.py
```diff
@@ -16,6 +18,9 @@ class Handler_ISBN(Handler):
     def inspect(self, citekey):
         import isbnlib
 
+        isbnlib.config.setthreadstimeout(seconds=timeout_seconds)
+        isbnlib.config.seturlopentimeout(seconds=timeout_seconds)
```
dhimmel, line 22: slight preference for us to define a helper like `import_isbnlib` such that we don't have to repeat this configuration 3 times.

dhimmel (commented):
other than comment looks good.

manubot/cite/isbn.py
xihh87: Is a helper preferable to using only a global import and config for the library at the top of the module?

This is an ISBN getting module anyway.

dhimmel: > Is a helper preferable to using only a global import and config for the library at the top of the module?

Usually not, although in this case let's assume we added the lazy importing of isbnlib for a reason and keep it that way.

dhimmel: @xihh87 did you want to switch this to use a helper so we can merge?

PR #346, opened 2022-11-06
Prepare 3.11 support and housekeeping

dhimmel:
https://docs.python.org/3/whatsnew/3.11.html was released on October 24, 2022!


dhimmel:
Blockers:

- [ ] ratelimiter is unmaintained and incompatible with 3.11 https://github.com/RazerM/ratelimiter/issues/10
- [ ] pysha3 incompatible with Python 3.11 and unmaintained https://github.com/opentimestamps/python-opentimestamps/pull/47

-

dhimmel:
Lot's of CI failures, but they mostly appear due to issues with the Zotero translation server (which I updated the source of during this PR at `translate.manubot.org`). I plan to merge this after one more look at the test failures.

Not official support for 3.11 until we can get rid of the pysha3 dependency, but this PR removes some of the other obstacles.

agitter:
Python 3.11 support will be great.  These changes all look reasonable.

Do we need to look into the Zotero translation server issues separately?  I didn't look at the test failures carefully but see a lot of failed tests.

dhimmel:
> Do we need to look into the Zotero translation server issues separately?

Yes, just opened https://github.com/zotero/translation-server/issues/140. I think I can downgrade our instance to get things working again. I think basically any tests that touch translation-server will fail until fixed.

Issue #82, opened 2018-12-04
Setting up a Zotero translation-server instance for citation metadata


dhimmel:
In #70, we discussed using some of Zotero's infrastructure for generating citation metadata. Specifically, the [translation-server](https://github.com/zotero/translation-server), which has [translators](https://github.com/zotero/translators) for many webpages.

I touched bases with the translation-server community in https://github.com/zotero/translation-server/issues/51 and there didn't seem to be any public instances. Therefore, we could either host a public instance for Manubot users or have Manubot users set up their own instances. While installing the nodejs package is not difficult, it does create a node dependency in an otherwise Python repository.

Therefore @dongbohu and I decided to spin up a translation-server at `https://translate.manubot.org`. It's currently hosted in a Google Cloud instance and we're still finalizing the setup. I'll let @dongbohu comment with the configuration details. While it'd be great to have a reproducible way to exactly configure this instance (perhaps via Terraform), we opted for the convenience of a traditional instantiation process.

dongbohu:
Here is a brief summary of the web server config (I will put more details later):
```
#!/bin/bash
# Update packages
sudo apt update
sudo apt upgrade -y

# Install nginx, nodejs, npm and supervisor
sudo apt install nginx supervisor nodejs npm

# Update npm to the latest version
sudo npm i npm -g

# Add new user "translate"
...

# Clone translation-server repo and set it up
...
# Configure supervisor to run translation-server at boot time
...

# Install certbot to take care of SSL
...

# Configure Nginx
...

# Reboot the box
```

Here is nginx config file:
```
# HTTP/HTTPS translation-server
server {
    if ($host != translate.manubot.org) {
        return 404; 
    }
    listen 80;
    listen 443 ssl;
    server_name translate.manubot.org;
    ssl_certificate /etc/letsencrypt/live/translate.manubot.org/fullchain.pem;   # managed by Certbot
    ssl_certificate_key /etc/letsencrypt/live/translate.manubot.org/privkey.pem; # managed by Certbot
    include /etc/letsencrypt/options-ssl-nginx.conf;                             # managed by Certbot
    ssl_dhparam /etc/letsencrypt/ssl-dhparams.pem;                               # managed by Certbot
    location / {
        proxy_pass http://127.0.0.1:1969;
        proxy_set_header X-Forwarded-Host $server_name;
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
    }
}
```

[...]

PR #347, opened 2022-11-06
use pyrate_limiter for rate limited requests

dhimmel:
throttler required using async which was too complicated Use pyrate_limiter instead.
https://github.com/vutran1710/PyrateLimiter

### DIFF
--- a/manubot/cite/pubmed.py
+++ b/manubot/cite/pubmed.py
@@ -1,4 +1,3 @@
-import asyncio
 import functools
 import json
 import logging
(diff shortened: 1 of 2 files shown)


dhimmel:
Replaces broken use of throttler in e2c5fd4f2d7c3e17f59936464941689c603b9f70 and 455842d990fd6f99a681f3924910da43e450ff00.

dhimmel:
Okay, I am pretty sure this is working from locally testing the following functions:

```
from manubot.cite.pubmed import get_pubmed_csl_item, get_pmid_for_doi
```

But the translate.manubot.org is now down related to https://github.com/manubot/manubot/issues/82#issuecomment-1304849387. So tests are taking forever and failing. Will merge with `[skip ci]`

PR #348, opened 2022-11-11
Test on Python 3.11

dhimmel:
follows https://github.com/manubot/manubot/pull/346

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -4,7 +4,7 @@ environment:
   matrix:
     - PYTHON: "C:\\Python37"
     - PYTHON: "C:\\Python37-x64"
-    - PYTHON: "C:\\Python311-x64"
+    - PYTHON: "C:\\Python310-x64"
 
 build: off
 
@@ -14,7 +14,6 @@ install:
   - "%CMD_IN_ENV% pip install .[webpage,dev] panflute==1.12.5"
 
 before_test:
-  - choco install openssl  # On Python 3.11, fix: Could not find module 'libeay32'
   - choco install pandoc --version 2.9.2
   - ps: $env:Path += ";C:\Program Files\Pandoc\"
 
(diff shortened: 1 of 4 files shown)


agitter (commented):
The changes you made here look good to me.  What are you thoughts about 3.11 on AppVeyor?  Your idea in 3bc0ea0525a0e6d700f51f780182dc80acfb5b07 is the only suggestion I would have had, and that didn't work.

dhimmel:
> What are you thoughts about 3.11 on AppVeyor?

I figured it's not worth spending much time fixing because:

1. we're testing windows with Python 3.11 on GitHub actions
2. it might resolve itself over time or at least workarounds might be better documented

dhimmel:
Wait now I see AppVeyor Windows with Python 3.10 is also failing with:

> FileNotFoundError: Could not find module 'libeay32' (or one of its dependencies).

So perhaps okay to upgrade AppVeyor to Python 3.11 and fix failure later.

agitter:
It makes sense to not spend more time debugging AppVeyor for the reasons you note above. Nice find with the linked issue.

A related question is whether we are still getting additional benefits from these AppVeyor tests given that we also test on Windows in GitHub Actions. They are different setup processes so can reveal different issues.  If maintaining the AppVeyor tests even became a major burden, we could consider dropping them though.

dhimmel:
> If maintaining the AppVeyor tests even became a major burden, we could consider dropping them though.

Yeah perhaps we reduce to a single OS + python version on appveyor and reduce to just the appveyor marked tests (removing integration below):

https://github.com/manubot/manubot/blob/053355653b41eae11fdcf4704e9e507dc636fa3b/.appveyor.yml#L21

agitter:
Did this AppVeyor error
> FileNotFoundError: Could not find module 'libeay32' (or one of its dependencies).

resolve itself?  I notice the AppVeyor test passed in #353

PR #349, opened 2022-11-12
manuscript metadata: explicit date

dhimmel:
helpful for https://github.com/manubot/rootstock/issues/283

Previously there was only a single date written to `variables.json` (the output manuscript metadata) and that was based on the generated timestamp of the manuscript.

Here we add support for a field in `metadata.yaml` named `date` to set an explicit publication date. The generated timestamp is still retained in the metadata and used to set some additional modified tags in the HTML header. This could be useful when you want to keep a fixed publication date on a manuscript but want to rebuild it to say fix a formatting issue or add new Manubot features.

### DIFF
--- a/manubot/process/tests/manuscripts/example/content/metadata.yaml
+++ b/manubot/process/tests/manuscripts/example/content/metadata.yaml
@@ -2,6 +2,7 @@
 title: "Example manuscript for testing"
 keywords:
   - manubot
+date: 2022-11-22
 authors:
   -
     github: dhimmel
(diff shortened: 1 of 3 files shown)


manubot/process/tests/manuscripts/example/content/metadata.yaml
```diff
@@ -2,6 +2,7 @@
 title: "Example manuscript for testing"
 keywords:
   - manubot
+date: 2022-11-22
```
dhimmel, line 5: Running the following tests:

```
pytest manubot/process/tests/test_process_command.py
```

Now produces the following output variables in `manubot/process/tests/manuscripts/example/output/variables.json`:

```json
  "manubot": {
    "date": "2022-11-22",
    "date_long": "November 22, 2022",
    "generated": "2022-11-12T08:50:02-05:00",
    "generated_date_long": "November 12, 2022",
```

agitter (approved):
This looks great.  Nice change.  The test failures appear unrelated.

We should update the rootstock usage with the expectations (ISO format) and purpose of this new `date` field in the metadata.

dhimmel:
> We should update the rootstock usage with the expectations (ISO format) and purpose of this new `date` field in the metadata.

Underway in https://github.com/manubot/rootstock/pull/478

Issue #354, opened 2022-11-27
Pubmed NCBI E-Utils API calls with `rettype=full` have an empty response


dhimmel:
CI Tests have been failing due to NCBI E-Utils EFetch Pubmed queries. URLs like https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=pubmed&id=21347133&rettype=full are returning status 200 (OKAY) but have an empty response causing XML parsing to fail. Removing `rettype=full` causes a complete response

dhimmel:
From https://support.nlm.nih.gov/?created=CAS-1023757-T0Z8S4

> Your new case has been created. Case number is CAS-1023757-T0Z8S4

dhimmel:
Got a reply from `NLM Support <nlm-support@nlm.nih.gov>`:

***

Hi,
 
Thanks for writing to us.
 
I don't believe rettype=full was an officially supported format. The full document is at:
https://www.ncbi.nlm.nih.gov/books/nbk25501 
 
The specific table for efetch is at:
https://www.ncbi.nlm.nih.gov/books/NBK25499/table/chapter4.T._valid_values_of__retmode_and/?report=objectonly 
 
The new update has some behavior change (which may enforce the rules more strictly). See this post for more information:
https://ncbiinsights.ncbi.nlm.nih.gov/2022/11/22/updated-pubmed-eutilities-live/ 
 
Regards,
 
Tao Tao, PhD
NCBI User Services
https://go.usa.gov/x647S

***

So looks like we should look into `rettype=xml` and if that differs in any way from `rettype=full`.

dhimmel:
> The [specific table](https://www.ncbi.nlm.nih.gov/books/NBK25499/table/chapter4.T._valid_values_of__retmode_and/?report=objectonly) for efetch

Copying from this table for `db = pubmed`:

> – Valid values of `&retmode` and `&rettype` for EFetch (null = empty string)
> 
> Record Type | &rettype | &retmode
> -- | -- | --
> XML | null | xml, default
> MEDLINE | medline | text
> PMID list | uilist | text
> Abstract | abstract | text

So `rettype=full` isn't a valid value anymore. We should set `retmode=xml`. I am not sure whether the old `rettype=full` is equivalent to the default response now.

PR #355, opened 2022-11-28
get_pubmed_csl_item: use rettype=xml

dhimmel:
closes https://github.com/manubot/manubot/issues/354

`rettype=full` is no longer supported.


agitter:
That's good news that the actual XML apparently didn't change as part of these NCBI E-Utils updates.

PR #358, opened 2023-01-22
Replace pre-commit checks with ruff

dhimmel:
[Ruff](https://github.com/charliermarsh/ruff#pycodestyle-e-w) is an "extremely fast Python linter, written in Rust". It replaces several of our existing tools including flake8, pyupgrade, and isort. It uses the same error codes when possible. Has been gaining popularity recently with rave reviews, so giving it a try.

### DIFF
--- a/manubot/cite/citekey.py
+++ b/manubot/cite/citekey.py
@@ -249,12 +249,15 @@ def shorten_citekey(standard_citekey: str) -> str:
 
 
 def citekey_to_csl_item(
-    citekey, prune=True, manual_refs={}, log_level: tp.Union[str, int] = "WARNING"
+    citekey, prune=True, manual_refs=None, log_level: tp.Union[str, int] = "WARNING"
 ):
     """
     Generate a CSL_Item for the input citekey.
     """
     from manubot import __version__ as manubot_version
+
+    if manual_refs is None:
+        manual_refs = {}
 
     # https://stackoverflow.com/a/35704430/4651668
     log_level = logging._checkLevel(log_level)
--- a/manubot/cite/tests/test_doi.py
+++ b/manubot/cite/tests/test_doi.py
@@ -28,7 +28,7 @@ def test_get_doi_csl_item_default():
     doi = "10.1101/142760"
     csl_item = get_doi_csl_item_default(doi)
     assert isinstance(csl_item, dict)
-    csl_item["publisher"] == "Cold Spring Harbor Laboratory"
+    assert csl_item["publisher"] == "Cold Spring Harbor Laboratory"
 
 
 def test_get_doi_csl_item_zotero():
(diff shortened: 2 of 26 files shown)


pyproject.toml
```diff
@@ -8,3 +8,24 @@ addopts = "--color=yes -rs"
...
+select = [
+    "B",  # flake8-bugbear
+    "C",  # flake8-comprehensions
+    "C90",  # mccabe
+    "E",  # pycodestyle errors
+    "F",  # pyflakes
+    "I",  # isort
+    "UP",  # pyupgrade
+    "W",  # pycode warnings
+]
```
dhimmel, line 28: Here is where we set rule codes to enable.

dhimmel:
Will merge now, since want to prevent conflicts with other PRs.

manubot/cite/citekey.py
```diff
@@ -249,13 +249,16 @@ def shorten_citekey(standard_citekey: str) -> str:
 
 
 def citekey_to_csl_item(
-    citekey, prune=True, manual_refs={}, log_level: tp.Union[str, int] = "WARNING"
+    citekey, prune=True, manual_refs=None, log_level: tp.Union[str, int] = "WARNING"
```
agitter, line 252: This is a good change. I recently was reminded of the unexpected behavior of this pattern.

setup.cfg
```diff
@@ -80,18 +80,3 @@ manubot =
     cite/curie/*.json
     process/header-includes-template.html
     webpage/redirect-template.html
-
-[flake8]
-ignore = E203, E266, E501, F541, W503
```
agitter, line 85: Are we choosing to modify the flake8 behavior by no longer ignoring these?

manubot/cite/tests/test_doi.py
```diff
@@ -42,7 +42,9 @@ def test_get_doi_csl_item_zotero():
     doi = "10.1038/ng.3834"
     csl_item = get_doi_csl_item_zotero(doi)
     assert isinstance(csl_item, dict)
-    csl_item["author"][9]["family"] == "GTEx Consortium"
+    # for an unknown reason, GTEx Consortium becomes the first author
```
agitter, line 45: This is very strange, and I don't understand why this reference has multiple inconsistent versions of the metadata.

agitter (commented):
I wanted to understand this pull request so I'm leaving a post-merge review. Everything looks good, especially ruff itself.

setup.cfg
dhimmel: According to https://lintlyci.github.io/Flake8Rules/, here are what those rules are:

```
E203	Whitespace before ':'
E266	Too many leading '#' for block comment
E501	Line too long (82 > 79 characters)
W503	Line break occurred before a binary operator
```

It's possible that we're no longer ignoring them if ruff implements and enforces them based on our selection. IIRC I just copied these rules from elsewhere, so probably not critical to ignore them.

PR #360, opened 2023-01-26
Add subcommand ai-revision (AI-assisted manuscript revision)

miltondp:
This PR adds the subcommand `ai-revision` for manubot. I've added initial instructions in the `README.md` file that will be updated after an initial review of the code and rest of the changes. In those instructions, you'll find how to run it.

A unit test fails when checking if the output in the `README.md` file is correct. The expected output splits a command parameter, and I think that would look weird.


manubot/ai_revision/ai_revision_command.py
```diff
@@ -0,0 +1,85 @@
+import logging
+
+
+def cli_process(args):
+    # set paths for content
+    content_dir = args.content_directory
+    if not content_dir.is_dir():
+        logging.warning(f"content directory does not exist: {content_dir}")
```
dhimmel, line 8: Does continuing make sense if there's no content directory?

```suggestion
        raise SystemExit(f"content directory is not a directory or does not exist: {content_dir}")
```

dhimmel (commented):
Looking great so far!

Can you merge main to get changes from https://github.com/manubot/manubot/pull/358? The pre-commit updates are significant.

It probably makes sense to transfer ownership of https://github.com/greenelab/manubot-ai-editor to the Manubot organization, since as a subcommand it should be maintained alongside everything else. At @cgreene's discretion.

manubot/ai_revision/ai_revision_command.py
```diff
@@ -0,0 +1,85 @@
...
+    if tmp_dir is None:
+        # if temporary directory is not given, generate one
+        import tempfile
+        from pathlib import Path
+
+        tmp_dir = tempfile.TemporaryDirectory()
+        tmp_dir = Path(tmp_dir.name)
+        logging.warning(f"output directory not specified, using: {tmp_dir}")
+
+    tmp_dir.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
miltondp: Yes, `tmp_dir.mkdir` is necessary. I'll take the other suggestions.

manubot/ai_revision/ai_revision_command.py
```diff
@@ -0,0 +1,85 @@
...
+
+    import shutil
+
+    from manubot_ai_editor import models
+    from manubot_ai_editor.editor import ManuscriptEditor
```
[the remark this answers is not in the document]
miltondp: Pefect!

manubot/ai_revision/ai_revision_command.py
```diff
@@ -0,0 +1,85 @@
...
+        model = module_class(
+            **model_kwargs,
+        )
+
+    # revise
+    me.revise_manuscript(tmp_dir, model, debug=True)
+
+    # move the revised manuscript back to the content folder
+    for f in tmp_dir.glob("*"):
+        shutil.move(f, content_dir / f.name)
```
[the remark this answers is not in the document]
miltondp: Yes, it's actually better. I ran into issues when the system temp dir was mounted on a different device. I had to switch to `shutil` because of that.

manubot/command.py
```diff
@@ -220,6 +221,43 @@ def add_subparser_webpage(subparsers):
...
+        nargs="+",
+        help="Keyword arguments for the revision model (--model-type), with format key=value.",
+    )
+    parser.add_argument(
+        "--temporary-directory",
+        type=pathlib.Path,
+        required=False,
+        help="Directory to temporarily output files generated by this script. "
+        "A temporary directory is created if this argument is not specified.",
+    )
```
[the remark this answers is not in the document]
miltondp: Yes, you're right; it's not really necessary. However, I always prefer to have the option to change this behavior. I added it when I ran into the problem of `/tmp` mounted on a different device and moving files. Now we are copying, so I will remove it.

miltondp:
> Looking great so far!

Thank you for your comments, @dhimmel! The code was much improved.

> Can you merge main to get changes from #358? The pre-commit updates are significant.

Done. I think I included all your suggestions, except the comment on `parse_kwargs` function. Not sure if there is something else to replace it.

> It probably makes sense to transfer ownership of https://github.com/greenelab/manubot-ai-editor to the Manubot organization, since as a subcommand it should be maintained alongside everything else. At @cgreene's discretion.

@cgreene already approved, so I can transfer ownership. The code and documentation in that repository need some improvement, especially the unit tests. Some use mock revision models to avoid calling OpenAI's API when possible, particularly when testing the editor component (if paragraphs are split correctly in the presence of equations, for instance, etc.). Other unit tests directly call the API, incurring costs. These would need some changes since they mainly test the prompts (for instance, if references are kept in a paragraph from the introduction), but unit tests are not the best way here since the LLM are stochastic.

README.md
```diff
@@ -302,6 +304,51 @@ options:
                         Set the logging level for stderr logging
 ```
 
+### AI-assisted academic authoring
+
+The `manubot ai-revision` command ... COMPLETE
+
+<!-- test codeblock contains output of `manubot ai-revision --help` -->
```
dhimmel, line 311: Based on test failures, looks like the help output here could be stale?

dhimmel (commented):
Looks great. I think we can merge once tests are passing. Will address any follow up comments from @agitter in subsequent PRs prior to a new release.

README.md
miltondp: I just [fixed it](https://github.com/manubot/manubot/pull/360/commits/b41cef8b298556946212d12ef481af0176d0a9ea). However, I think the expected, split output adds a new line in the middle of a parameter name (`--model-type`) which doesn't look great. But it's a very minor detail.

miltondp:
> Looks great. I think we can merge once tests are passing. Will address any follow up comments from @agitter in subsequent PRs prior to a new release.

I need to fix something else in `README.md` and it would be ready.

README.md
miltondp: The complete instructions to use the tool are in manubot/rootstock#484. A basic setup (such as setting `OPENAI_API_KEY` is missing here). Maybe we should add a link to rootstock's `USAGE.md`?

dhimmel: We could play with https://docs.python.org/3/library/argparse.html#formatter-class but probably not worth the hassle.

manubot/command.py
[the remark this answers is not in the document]
agitter: I've run into problems running Python code that uses default tmp directories in Singularity containers in the mode with a non-writeable filesystem. There are usually workarounds for that, but having this argument would be more convenient.

That said, I don't have a strong opinion either way.

manubot/ai_revision/ai_revision_command.py
```diff
@@ -0,0 +1,79 @@
...
+            **model_kwargs,
+        )
+
+    # revise
+    me.revise_manuscript(tmp_dir, model, debug=True)
+
+    # copy the revised manuscript back to the content folder
+    for f in tmp_dir.glob("*"):
+        shutil.copyfile(f, content_dir / f.name)
+
```
agitter, line 48: Does the code need to delete `tmp_dir` here? https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp says 
> The user of mkdtemp() is responsible for deleting the temporary directory and its contents when done with it.

Everything else looks good to me.

dhimmel: Linux deletes temporary files on reboot. I guess windows might not delete them by default. I always thought it was the OSes responsibility to cleanup the temporary directories but I might be wrong.

miltondp: It should not be a problem. The folder is in the system's temporary directory, which will be deleted at some point. And `.md` files there are expected to be small, so there should be no concerns about disk usage.

If we want to delete it, we could put the code inside a `try` block and ignore any errors.

agitter: Testing on Windows, my default temp directory is
```
>>> tempfile.gettempdir()
'C:\\Users\\agitter\\AppData\\Local\\Temp'
```

I see files there from 2021, which is consistent with discussion [here](https://superuser.com/questions/571901/is-it-safe-to-delete-appdata-local-temp) about running the Disk Cleanup utility to delete temporary files.

On the other hand, the COVID-19 review is the longest Manubot manuscript. Its cumulative Markdown content is still only 1.1. MB, so leaving copies of that in the temp directory isn't going to have any practical consequence. I'm fine leaving it as is.

PR #369, opened 2024-02-17
FIX: Test fails for no reason.

xihh87:
manubot/pandoc/tests/test_cite_filter.py::test_cite_pandoc_filter fails on main.

The website [1] states in it's metadata that the article dates from 2016-11-02.

<meta name="citation_publication_date" content="2016/11/02"/>

Most likely, because that is the time for the first submission [2].

Having the test fails because it does not gives the date stated on the website does not good, as the tool is working as intended.

If someone cares about this date, either make a custom local citation or ask the admins of the website to update the metadata for the article.

[1]: https://openreview.net/forum?id=HkwoSDPgg
[2]: https://openreview.net/revisions?id=HkwoSDPgg


dhimmel (approved):
Thanks for the CI fix.

These errors arise because the expected test output has frozen citation metadata at some past point, but the tests re-generate expected output as of the latest citation metadata.

Issue #379, opened 2025-11-06
Use GitHub Dependabot to help maintain dependencies


d33bs:
This project might benefit from automated dependency updates from GitHub Dependabot. If the project team feels good about this, I'd be interested in exploring adding this to the project through a contribution. Thanks!

dhimmel:
Are you envisioning dependabot being used for this python package, whose dependencies are as unpinned as possible for wide compatibility:

https://github.com/manubot/manubot/blob/2914d720dfe02d2bda8781f8f784bc14d461e2e3/setup.cfg#L41-L55

Or for rootstock which does [pin](https://github.com/manubot/rootstock/blob/ca748bf4e80a830d7bee69a20b1fae18e028d9e6/build/environment.yml) dependencies?

For the manubot package, fixing CI failures is very helpful and likely a prerequisite to automated dependency updates. But then again, the python package tries to be permissive with dependency versioning.

Rootstock tries to pin everything, but doesn't have much of a test suite. Hence the conundrum.

I've been using Renovate recently. Another option to consider besides Dependabot.

d33bs:
Thanks @dhimmel ! I didn't realize the philosophy for this repo and rootstock - thanks for this clarity.

I think we could start with dependency updates for GH actions (e.g. `actions/checkout`, `actions/setup-python`, which are out of date).

Renovate sounds cool! Do you know if it can also check for GH actions dependencies?

---

Outside of this issue's focus, it could be helpful to explore how the project maintains dependencies more generally.

I feel there are benefits to using lockfiles and being specific about what is tested. That said, a `pip freeze` requirements file could be a bare minimum. Alternatively, I'd vote for using `uv` or `poetry` (which have their own lockfiles) if a larger upgrade sounds okay to you.

dhimmel:
> I think we could start with dependency updates for GH actions (e.g. actions/checkout, actions/setup-python, which are out of date).

No objections here, it's good to update action dependencies. Doing this manually is likely easiest. I haven't used Dependabot, maybe it is easy to set up because it is part of github. If so, worth a try. Renovate is more full featured, but likely more to configure and requires authorizing an external app (or adding another CI workflow).

> Outside of this issue's focus, it could be helpful to explore how the project maintains dependencies more generally. I feel there are benefits to using lockfiles and being specific about what is tested

[...]

PR #380, opened 2025-11-10
Add GitHub Actions Dependabot configuration

d33bs:
This PR adds a GitHub actions Dependabot configuration which will automatically check and issue updates for actions used by the project. On merge, this will run for the first time, likely issuing PR's for stale actions which need an update.

References #379


dhimmel (approved):
Test failures are unrelated.

I haven't evaluated whether dependent runs with this config.

Do we need to merge it first to see?

PR #383, opened 2025-11-11
Update tests to avoid unnecessary failures

d33bs:
I noticed in #380 that there were a few tests which didn't pass (despite being irrelevant to the changes). I wanted to submit a PR to help address these where appropriate and within reason to help avoid any unnecessary failures related to external or internal data status or formatting.


d33bs:
Thanks @dhimmel !

PR #385, opened 2025-11-29
Add recent Python versions to tests matrix

d33bs:
This PR adds recent versions of Python to the tests matrix for CI, expanding the existing coverage. I opted to leave 3.14 out for now due to how new it is currently (and likelihood for challenges unrelated to the code here).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,6 +18,8 @@ jobs:
         python-version:
         - "3.10"
         - "3.11"
+        - "3.12"
+        - "3.13"
         pandoc-version:
         - "2.9.2"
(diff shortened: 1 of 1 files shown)


agitter (approved):
Makes sense to me. I'll let @dhimmel check it out too.

.github/workflows/test.yml
```diff
@@ -18,6 +18,8 @@ jobs:
         python-version:
         - "3.10"
         - "3.11"
+        - "3.12"
```
dhimmel, line 21: consider just testing on the min and max supported python like 3.10 and 3.13 or 3.14 to reduce load on the API endpoints we hit.

